PR #635, opened 2026-07-16
Add ShapeIndicator component

### LINKED ISSUE #487: [Parity Check] Investigate ShapeIndicator component

**Component**: ShapeIndicator
**Commit**: 1110000

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/ShapeIndicator
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-shapeindicator--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/1110000e8374c35b7e040b09a7e292b2227f331d/packages/react/src/components/ShapeIndicator)

---
*Auto-generated by parity-check script on 2026-07-14T19:57:45.994Z*

patricklx:
## Summary
- Adds the `ShapeIndicator` component (React's `preview__ShapeIndicator`), which was missing from the Ember implementation (parity gap for #487)
- Mirrors the React API: `@kind` (11 status kinds), `@label`, `@textSize` (12/14), and `@compact` (shape-only with an accessible tooltip label)
- `@compact` uses `ember-attacher`'s `<AttachTooltip>` for the hover/focus tooltip, since that's already how tooltips are done elsewhere in this codebase (e.g. `copy-button.gts`) and there's no dedicated Tooltip/Popover component here yet. React's `align`/`autoAlign` props are intentionally omitted since they only tune this positioning, which `ember-attacher` already handles with sensible defaults.
- Exported from `carbon-components-ember/components`, documented in docs-app, and covered by integration tests (including style snapshots for light/dark themes)
- Fixes a pre-existing bug in `render-svg-part.ts`: `Array#join()` was called with no separator, defaulting to `,`, which injected a literal comma text node into the markup of any multi-path icon (e.g. the two-path `Critical` icon used for `kind="failed"`). This was visible immediately in ShapeIndicator's default rendering, so it's fixed as part of this PR.

Closes #487

## Test plan
- [x] `pnpm build` in `carbon-components-ember` succeeds
- [x] `pnpm test:ember` in `test-app` passes (23/23), including new `Integration | Component | ShapeIndicator` tests
- [x] Verified rendering in `docs-app` matches the React Storybook default/14px stories (colors, icons, layout)


patricklx:
Compact mode shows a button with icon inside, it should not display a button

patricklx:
Addressed the review feedback:

- **Compact mode no longer renders a `<button>`.** There's no `cds--shape-indicator__button` styling in `@carbon/styles`, so the native button chrome was showing through unstyled. Since the compact indicator isn't an interactive control (it doesn't do anything on click), it now renders as a `<span tabindex="0">` instead — still focusable/hoverable so the `AttachTooltip` works via mouse and keyboard, but without the button semantics/appearance. (`carbon-components-ember/src/components/shape-indicator.gts`)

Verified:
- `pnpm build` in `carbon-components-ember` succeeds
- `pnpm test:ember --filter ShapeIndicator` in `test-app` passes (6/6), including the compact-mode accessibility test
- `pnpm lint` passes with no new warnings/errors

patricklx:
Tooltip does not fully cover the text
Screenshot_20260718_104247_Chrome

patricklx:
Fixed the tooltip clipping issue.

**Root cause:** it's a pre-existing bug in `ember-attacher`'s own CSS, exposed by ShapeIndicator's compact mode because its trigger (a single 16px icon) and label are both unusually small. On screens ≤450px, ember-attacher applies `.ember-attacher-tooltip { max-width: calc(100% - 42px); }`, assuming `100%` resolves to the viewport width. But ember-attacher actually positions tooltips inside a `width: max-content` wrapper that's sized to the tooltip's own content — so for a short tooltip like "Failed", `100%` resolves to that small intrinsic width (~75px) instead of the viewport, and the calc clamps the bubble down to ~33px, clipping the label ("Failed" rendered as "Fa").

I reproduced this precisely against the deployed docs preview for this PR at a mobile viewport, confirming the clamped width (33.265625px) exactly matched `75.265625px - 42px` from the broken calc.

**Fix:** added a scoped override in `carbon-components-ember/src/styles/index.scss` (same file already used for an `ember-basic-dropdown` CSS fix) that replaces the broken rule with `max-width: min(90vw, 400px)`, using `vw` units so it's always relative to the actual viewport regardless of the tooltip's containing block.

Verified:
- `pnpm build` in `carbon-components-ember` succeeds and the compiled CSS bundle includes the override with correct cascade precedence over ember-attacher's rule
- `pnpm test:ember --filter ShapeIndicator` in `test-app` passes (6/6)
- `pnpm lint` passes with no new warnings/errors

PR #637, opened 2026-07-18
Rename UiShell export to UIShell for parity with React

patricklx:
## Summary
- The Ember `UIShell` (Header + Nav + SideNav composition, `carbon-components-ember/src/components/ui-shell.gts`) already existed but was exported as `UiShell` instead of `UIShell`, which is the exact React casing (`packages/react/src/components/UIShell`).
- The casing mismatch caused the parity checker to treat `UIShell` as missing on the Ember side while separately listing `UiShell` as an "Ember-specific" component — it's actually the same component under a mismatched name.
- Renamed the export in `index.ts` and the class/interface names throughout `ui-shell.gts` and its sub-components (`-header.gts`, `-nav.gts`, `-nav/-item.gts`, `-sidenav.gts`) to `UIShell`/`UIShellHeader`/`UIShellNav`/`UIShellNavItem` to match React.
- Added a smoke test (`test-app/tests/components/ui-shell-test.gts`) since no test previously existed for this component.

This is a naming-only alignment; no behavior changed.

Fixes #507

## Test plan
- [x] `pnpm build` (glint type-check + rollup) succeeds with no errors
- [x] Verified no remaining `UiShell` references in `carbon-components-ember/src`
- [ ] Full browser-based qunit suite (`pnpm test:ember` in `test-app`) — could not be executed in this sandbox: the dev server's first-compile of `vendor.css` alone takes ~36s, exceeding the test runner's hardcoded 30s Playwright navigation timeout. This is a pre-existing environment/performance limitation, unrelated to this change (reproduces identically on a clean build). Please run `pnpm test:ember` in CI/locally to confirm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #639, opened 2026-07-18
fix no live preview

patricklx:
There is no live preview for uishell. Needs to be fixed

### DIFF
--- a/.github/workflows/templates/agent-prompt.md
+++ b/.github/workflows/templates/agent-prompt.md
@@ -50,6 +50,11 @@ Only exclude a component when there's a genuine reason (React-only concept/utili
 
 #### Path A: Component Exists - Fix Discrepancies
 
+If comparison finds the Ember component already matches React with no
+discrepancies, that's a valid outcome of Path A too — skip straight to
+Phase 3/4 with no code changes. Don't skip Phase 4: you still mark the
+component synced and push that update (see Phase 4, step 3).
+
 1. **Check for a Naming Mismatch**
    - If the component already exists in Ember under a different name (e.g. `carbon-components-ember/src/components/` has an equivalent component with a different export name), rename/align it to match the React name instead of treating it as missing
    - Update the export in `index.ts`, the file name, and any references/docs/tests accordingly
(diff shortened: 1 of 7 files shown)


patricklx:
There is no live preview for uishell in docs. Needs to be fixed.
The markdown with live preview Mark is not working 

Uncaught (in promise) TypeError: Cannot convert undefined or null to object

Maybe some missing component that is not re-exported or not configured correctly in docs app?

patricklx:
## Fix applied ✅

**Root cause:** the ui-shell live preview imports `Notification` and `UserAvatar` from `carbon-components-ember/icons`, but the docs app's kolay `resolve` map (`docs-app/app/routes/application.ts`) only exposed six other icons. Both imports therefore resolved to `undefined`, and the `Icon` component threw `TypeError: Cannot convert undefined or null to object` when iterating `icon.attrs` — which killed the entire live preview.

**Changes in 1f4113d4:**
- `docs-app/app/routes/application.ts` — added `Notification` and `UserAvatar` to the `carbon-components-ember/icons` resolve map.
- `carbon-components-ember/scripts/create-files.mjs` — special-case the `UIShell` casing so regenerating `src/components/index.ts` doesn't revert the `UiShell` → `UIShell` rename from #637 (which would break the preview's import again).
- `docs-app/public/docs/2-components/ui-shell.md` — reverted the stray blank line added in the earlier attempt.

**Verified end-to-end** with a Playwright check against a local docs build of `/2-components/ui-shell`: the preview now renders the shell header (`cds--header`), the side nav, and both global action buttons, with no console errors.

patricklx:
The ui shell live preview covers thr entire page. Should be only inside a contained box.
Also, i cannot click on e.g. Category 1. Nothing happens, it doesn't extend

patricklx:
## Follow-up fixes applied ✅

Addressed the latest feedback (preview overlaying the page + "Category 1" not expanding) in 9c01ef62:

**1. Preview is now inside a contained box**
Carbon's `cds--header` and `cds--side-nav` are `position: fixed`, so the demo escaped the preview area and covered the whole docs page. The demo in `docs-app/public/docs/2-components/ui-shell.md` now wraps the shell in a bordered, fixed-height wrapper with a `transform`, which makes the wrapper the containing block for the fixed-position header and side nav — the shell renders entirely inside the box.

**2. "Category 1" now expands/collapses on click**
Two causes:
- The demo passed `@submenus={{array}}` (empty), so the item rendered as a plain link wired to a noop. It now has real sub-links (`Sub-link 1`, `Sub-link 2`).
- A real bug in `ui-shell/-sidenav/-menu.gts`: the expanded flag was stored in an untracked plain object (via the `new-obj` helper + a `toggle` function misused with `fn`), so clicking never re-rendered — and rendering the submenu branch actually crashed with `You must pass a function as the fn helper's first argument`. The component now uses tracked state (`@tracked expanded` + a `toggleExpanded` action).

**Tests:** added an integration test for submenu expand/collapse; all 6 UIShell tests pass.

**Verified in the browser** with a Playwright check against a local docs run of `/2-components/ui-shell`: the header renders inside the preview box (not across the viewport), Category 1 expands to show its sub-links and collapses again, with no console errors.

PR #649, opened 2026-07-21
feat: rename Radio/RadioGroup to RadioButton/RadioButtonGroup for React parity

patricklx:
## Summary

- The Ember RadioButton implementation already existed but was exported as `Radio`/`RadioGroup` instead of `RadioButton`/`RadioButtonGroup`, and its API diverged noticeably from Carbon React's `RadioButton`.
- Renamed `radio.gts` → `radio-button.gts` and `radio/group.gts` → `radio-button/group.gts`; updated the exports in `index.ts` and `package.json`.
- Aligned args with React: `isChecked`/`isDefault` → `checked`/`defaultChecked`, added `hideLabel`, `labelPosition`, `name`, `required`, `readOnly`, `id`, `onClick`.
- Reworked group selection to be value-based (`valueSelected`/`defaultSelected`, comparing `@value`) instead of tracking component instances, and renamed the group's `onSelect` → `onChange` with a `(value, name, event)` signature matching React's `RadioButtonGroup`.
- `RadioButtonGroup` now also accepts a `@legendText` string arg (in addition to the existing `heading` named block) and cascades `disabled` to its children.
- Updated the docs page (`radio.md` → `radio-button.md`) and the `Radio`/`RadioGroup` references in `accordion.md`/`checkbox.md`.
- Added integration tests covering rendering, `checked`/`defaultChecked`, `disabled` (including group cascade), `hideLabel`, `labelPosition`, and `onChange` for both `RadioButton` and `RadioButtonGroup`.
- Ran the parity-check script, which also picked up several other components (`Slider`, `Stack`, `ShapeIndicator`, `Text`, `Theme`, `UIShell`, etc.) that were already implemented in `index.ts` but not yet reflected in `.parity-check-data.json`/`PARITY_REPORT.md`.

Note: `pnpm build:types` currently fails on `main` independent of this change — `index.ts` exports `Slider`, `SliderSkeleton`, and `Stack` from files that don't exist in this branch (they exist only on the unmerged `pr-644-inspect` branch for issue #488). I verified in isolation (temporarily removing those three lines) that the build and the full `test-app` suite (74/74) pass cleanly with this change.

Closes #480

## Test plan

- [x] `cd carbon-components-ember && pnpm build` (verified in isolation without the pre-existing broken Slider/Stack exports; those failures are unrelated to this change)
- [x] `cd test-app && pnpm test` — 74/74 passing
- [x] `npx eslint` on the touched `.gts` files — no errors

[...]

PR #656, opened 2026-07-21
feat: rename Menu to OverflowMenu for React parity

### LINKED ISSUE #469: [Parity Check] Investigate OverflowMenu component

**Component**: OverflowMenu
**Commit**: 1110000

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/OverflowMenu
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/1110000e8374c35b7e040b09a7e292b2227f331d/packages/react/src/components/OverflowMenu)

---
*Auto-generated by parity-check script on 2026-07-14T19:57:14.313Z*

patricklx:
## Summary
- The Ember `Menu` component (`menu.gts`) already implemented Carbon React's `OverflowMenu` — it uses the `cds--overflow-menu` / `cds--overflow-menu-options` CSS classes and the button + floating-options-list structure that matches React's `OverflowMenu`, not React's separate `Menu` (context menu) component.
- The parity-check bot flagged `OverflowMenu` as missing purely because of this naming mismatch (matches the precedent set by the Radio→RadioButton rename in #649).
- Renames `menu.gts` → `overflow-menu.gts` and `menu/-item.gts` → `overflow-menu/-item.gts`, updates the internal `DataTable` row-actions menu to use the new import paths, regenerates `index.ts`/`package.json` exports, renames the docs page, and adds integration tests.
- Note: after this rename, React's `Menu` (context menu) now correctly shows as missing/unimplemented in the parity report, since the Ember component never actually matched that component's API — it only accidentally shared the name.

Closes #469

## Test plan
- [x] `pnpm build` succeeds in `carbon-components-ember`
- [x] `pnpm test` passes in `test-app` (114/114, including 6 new OverflowMenu tests)
- [x] Manually reviewed generated `index.ts`/`package.json` export diffs


patricklx:
Overflow menu closes immediately after click menu button, it only happens on mobile. works find on desktop

patricklx:
Investigated the "closes immediately on click, only on mobile" report.

**What I checked:** walked through `ember-basic-dropdown`'s trigger/content touch handling (it dispatches a synthetic `click` via `setTimeout(0)` after `touchend`, guarded by a `toggleIsBeingHandledByTouchEvents` flag meant to prevent a double-toggle). Then reproduced with a real touch tap — Chromium mobile emulation (iPhone 13 profile, `hasTouch`/`isMobile`), `OverflowMenu` with `@tooltip` set, in an isolated render (not the docs-app demo). Result: the menu **stays open** correctly after a single tap; the short-circuit works as designed.

**What I also found:** the docs-app's live/editable demo (glimdown) on this same page never stabilizes locally — its rendered DOM keeps changing every ~1s (ties back to the "Recrawled this watch 363 times" file-watcher warning from the dev server). When I taped the trigger there, the demo's live recompile tore down and rebuilt the whole preview shortly after, which looks exactly like "opens then closes immediately." That's a pre-existing docs-app dev-server issue, unrelated to this PR's component code.

So: no bug found in the component itself under a clean, isolated touch tap. Two possibilities for what you saw:
1. It was the docs-app live-demo instability above — worth re-testing against the deployed preview (production build, no live recompile) rather than a local dev server.
2. It's a WebKit/Safari-specific touch→click timing quirk that Chromium's touch emulation doesn't reproduce (I don't have a WebKit-based browser available here to test).

Added `should stay open after a touch tap on the trigger (mobile)` as a permanent regression test (pushed in 279ad765) covering the touch/click short-circuit path, so this'll be caught in CI if it regresses. If you can still reproduce this on a real device against the docs preview, more detail (device/browser, whether it's the docs demo specifically or a plain render) would help narrow it further.

patricklx:
Let's remove the hard coded eventType='click' .
Make it configurable, do not set a default value for it

patricklx:
Addressed the review feedback: removed the hardcoded `@eventType='click'` on the dropdown trigger and exposed it as a new `@eventType?: 'click' | 'mousedown'` arg on `OverflowMenu`, left undefined by default (no forced default) so `ember-basic-dropdown`'s own internal `'click'` default takes over, as requested.

- `carbon-components-ember/src/components/overflow-menu.gts`: signature + template change (with a `@glint-expect-error` since the trigger's glint types mark `eventType` as required even though the runtime treats it as optional).
- Added a regression test asserting `@eventType='mousedown'` actually opens the menu on a `mousedown` event.
- `pnpm build` and `pnpm glint` both pass cleanly. Local Playwright-based `test-app` integration test run hit a pre-existing harness issue unrelated to this change (`run-tests.mjs` throws on a non-JSON `[HARNESS] ci=true` stdout line before any test executes) so I couldn't get a live pass/fail from that specific runner locally — relying on the build/typecheck plus the isolated logic review here.

Re-ran `--mark-synced OverflowMenu` so the parity data reflects this update.

PR #655, opened 2026-07-21
feat: rename SearchInput to Search for React parity

patricklx:
Closes #483

`Search` already existed in Ember but was exported as `SearchInput`, diverging from Carbon React's naming. This renames the component/export to `Search` and updates all internal references (`data-table`, `list`) and docs.

## Summary
- Renamed `search-input.gts` → `search.gts`, exported as `Search` (was `SearchInput`)
- Updated `data-table/-search-input.gts` and `list.gts` to import from the new path
- Updated docs at `docs-app/public/docs/2-components/search.md`
- Added integration tests covering rendering, size, disabled state, placeholder/value, onChange, and clear button behavior
- Marked `Search` as synced in `.parity-check-data.json` / `PARITY_REPORT.md`

## Test plan
- [x] `pnpm build` succeeds in `carbon-components-ember`
- [x] `pnpm run test:ember` passes (118/118, including 10 new Search tests)
- [x] `pnpm run lint` passes (0 errors)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #659, opened 2026-07-22
feat: implement TreeView component

### LINKED ISSUE #506: [Parity Check] Investigate TreeView component

**Component**: TreeView
**Commit**: 1110000

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/TreeView
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-treeview--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/1110000e8374c35b7e040b09a7e292b2227f331d/packages/react/src/components/TreeView)

---
*Auto-generated by parity-check script on 2026-07-14T19:58:20.362Z*

patricklx:
## Summary
- Adds a new `TreeView` component for parity with Carbon React's `TreeView`/`TreeNode`.
- Renders an accessible tree (`role="tree"`, `role="treeitem"`, `role="group"`) and yields a bound `TreeNode` component that can be nested arbitrarily deep to build out the hierarchy — usage mirrors the `Accordion`/`Item` yield pattern already used in this codebase.
- Supports `@label`/`@hideLabel`, single-select or `@multiselect` (Cmd/Ctrl+click for additional nodes), `@onSelect`/`@onActivate` callbacks, `@size` ('xs'/'sm'), and per-node `@id`/`@disabled`/`@isExpanded`.
- Expand/collapse via a toggle caret (mirrors Carbon's `caret--right` icon, rotated when expanded) plus `ArrowLeft`/`ArrowRight` keyboard support; `Enter`/`Space` selects the focused node.
- Uses `cds--tree`/`cds--tree-node`/`cds--tree-parent-node`/`cds--tree-node__children` etc. classes so the existing `@carbon/styles` CSS (already bundled) styles it correctly.

Simplifications vs. upstream React (documented as follow-ups, not blockers): no `renderIcon`/custom per-node icon slot, no `href` link-node variant, and no full roving-tabindex arrow-key traversal between sibling nodes (each enabled node is independently focusable via Tab instead).

Closes #506

## Test plan
- [x] `pnpm build` (types + rollup) succeeds
- [x] `pnpm lint` passes for the new file (pre-existing warnings elsewhere untouched)
- [x] New tests in `test-app/tests/components/tree-view-test.gts` cover: label/hideLabel, roles, expand/collapse + `@isExpanded`, click-to-select, disabled nodes, and `@multiselect` — all passing
- [x] White/dark theme style snapshot tests added


patricklx:
the tree view in docs app does not collpase when clicking items (the arrow changes though)

patricklx:
Thanks for catching that — confirmed and fixed in 99ba3ca0.

**Root cause:** the collapse toggle was adding/removing a `cds--tree-node__children--hidden` class, but that class doesn't exist anywhere in `@carbon/styles` — so it had no effect. That's why the caret rotated (`cds--tree-parent-node__toggle-icon--expanded` *is* a real class) but the subtree never actually hid.

**Fix:** switched to the class Carbon actually ships, `cds--tree-node--hidden` (`display: none`), applied to the `cds--tree-node__children` wrapper `<ul>`.

One wrinkle worth noting: I initially tried applying `cds--tree-node--hidden` directly to each collapsed child `<li>` (closer to how upstream React marks up hidden nodes), but that collides with `@carbon/styles`' own cascade — `.cds--tree-leaf-node { display: flex; }` is declared later in the stylesheet than `.cds--tree-node--hidden { display: none; }`, and with equal specificity the later rule wins, so leaf nodes stayed visible. Hiding the wrapping `<ul>` instead avoids that collision entirely (the `<ul>` doesn't carry any conflicting `display` rule), so it's both correct and a smaller diff.

Also added a test that loads the real Carbon CSS and asserts on actual visibility (`isVisible`/`isNotVisible`), not just class presence — that's the test that would have caught this originally, since the existing test only checked for the (bogus) class name rather than the rendered effect.

Verified manually in the docs app (dev server + browser automation): collapsing "Cloud computing" now correctly hides its children instead of just rotating the caret.

All tests pass (174/174), lint clean, build succeeds.

patricklx:
Found and fixed a data-integrity issue in `.parity-check-data.json` that CI/tests don't catch: TreeView was added to the `ember` component list and `componentMetadata`, but was left in `components.missing` and never added to `components.implemented` — unlike the equivalent update in PR #660 (Tooltip). This meant the parity tracker would still consider TreeView unimplemented after merge.

Fixed in 90a0f249 (moved `TreeView` from `missing` to `implemented`, alphabetically sorted). Verified with a key-set diff against `origin/main` (no other unexpected drift), and re-ran `build:carbon`, `lint`, and `test:ember` (174/174 passing).

patricklx:
when selecting an item, it should select the whole row, not at start of text item. the arrow should point down when expanded, and point right when closed. review all examples in carbon react story, and also add them to the docs

patricklx:
Addressed the latest round of feedback in 29902f79:

**Row selection now highlights the whole row.** The selected/hover background on `.cds--tree-node__label` was only extending from where the label text starts (leaving the left "gutter" — where a toggle caret would sit — unhighlighted), instead of spanning the full row. Root cause: our markup was missing the per-node negative-margin/padding trick upstream Carbon React applies (`TreeNode.tsx`'s `calcOffset`), which is what makes the highlight flush with the tree's true left edge regardless of nesting depth or leaf/parent type. Added a `depth` arg that's threaded automatically as nodes are yielded, and used it to compute the same offset on `.cds--tree-node__label` via `margin-inline-start: -Xrem; padding-inline-start: Xrem`. This also fixes a related gap: nested rows weren't visually indented per level at all before (indentation was previously capped at a fixed 1rem/2.5rem regardless of depth); it now grows correctly with nesting depth, matching upstream. Verified by rendering the exact markup with real `@carbon/styles` CSS in a headless browser before/after — screenshots confirmed the highlighted rectangle now touches the left edge at every depth.

**Toggle caret now points right when collapsed, down when expanded.** We were using the `caret--right` SVG path directly combined with `@carbon/styles`' `rotate(-90deg)`/`rotate(0)` rules — but those rotation rules are meant to be applied to the `caret--down` icon (which is what upstream `TreeNode.tsx` actually renders). Right-path + those rotations produced up/right instead of right/down. Swapped to the real `caret--down` path (`M24 12 16 22 8 12z`); same CSS classes now produce the correct right→down transition.

**Reviewed the Carbon React story** (`Treeview.stories.js` + `TreeNode.tsx` from carbon-design-system/carbon) against our implementation. Confirmed `renderIcon` and `href` (link-node variant) remain out of scope as already documented. Added a live docs example exercising `@multiselect`, `@size`, and `@onSelect` together (previously only mentioned in prose, not demonstrated).

Verification done locally: `pnpm build:carbon` (types + rollup) succeeds, `pnpm lint` is clean (0 errors — fixed a `style-concatenation` lint error along the way by moving the dynamic style into `htmlSafe` getters, matching the existing pattern in `code-snippet.gts`), and `glint` typechecks the addon and the new test code with no new errors. Also added two new tests (toggle caret path/class assertions, and a real-CSS geometry test asserting the label background's bounding rect is flush with its row at multiple depths).

Note: I wasn't able to run the local Ember/Playwright test suite (`test:ember` or `ember test`) — it fails with a generic `vite build failed` error that I confirmed reproduces identically on a clean checkout of this branch with none of my changes applied, so it's pre-existing local environment breakage, not something introduced here. Verification was instead done via `build:carbon` + `lint` + `glint` + a standalone headless-browser render of the actual markup/CSS to visually confirm both fixes.

patricklx:
also add to the docs "controlled expansion" , "with icons" and "disabled state"

patricklx:
Addressed the latest round of feedback in 5b008359:

**`@icon` support.** Added an `@icon` arg to `TreeNode` (any icon component from `carbon-components-ember/icons`, e.g. `Folder`/`Document`), rendered before the label via `<@icon @svgClass='cds--tree-node__icon' />` (mirrors the existing `sidenav`'s `@icon` pattern in this codebase). Also adds the real `cds--tree-node--with-icon` modifier class Carbon ships for icon-adjusted spacing.

**Controlled expansion.** Added an `@onToggle` callback. `@isExpanded` now behaves like this: uncontrolled by default (only seeds the initial state, exactly as before — clicking the caret still toggles it freely), but if `@onToggle` is also passed, the node becomes fully controlled and always reflects `@isExpanded`, with `@onToggle` reporting the intended next state back to the consumer. This was deliberately *not* "just always prefer `@isExpanded` if defined" — that naive version breaks the plain `@isExpanded={{true}}` initial-state-only usage already shipped (it would permanently lock the node open since a literal arg is never `undefined`). I caught this in local testing (see below) before it went out.

**Docs additions:**
- "With icons" — `Folder` on parent nodes, `Document` on leaves.
- "Controlled expansion" — a button that expands/collapses "Cloud computing" externally via `@isExpanded`/`@onToggle`.
- "Disabled state" — disabled parent and leaf nodes, with a note that a disabled parent's toggle caret stays usable so descendants stay reachable.

**Verification:** this time I *was* able to run the full local suite end-to-end (`test:ember` via `run-tests.mjs`, headless Chromium) — 198/198 passing, including 4 new tests (icon rendering + with-icon class, uncontrolled-toggle-still-free-after-initial-@isExpanded, and controlled-expansion-via-@onToggle). Also `pnpm build:carbon` (types + rollup) and `pnpm lint` (0 errors) are clean.

One thing worth flagging from the process: my first pass at the controlled-expansion getter (`this.args.isExpanded ?? this.uncontrolledExpanded`) looked right but was subtly broken — since `??` only falls through on `null`/`undefined`, any explicitly-passed `@isExpanded` (`true` *or* `false`) would permanently override clicks, silently breaking the existing "initial state only" behavior for every tree in the docs that passes a literal `@isExpanded`. The full local test run caught it (new tests failed) before I rebuilt/pushed, so it's fixed in this commit — mentioning it since it's the kind of regression that's easy to miss without exercising the actual click path.

patricklx:
Fixed merge conflicts and a failing required DCO check:

- Merged latest `main` (TextInput component landed) — conflict was only in `.parity-check-data.json`, both `TreeView` and `TextInput` metadata entries kept.
- The tip commit (`5b008359`, icon/controlled-expansion feature) had a `Co-Authored-By` trailer but was missing the `Signed-off-by` trailer that DCO requires — every other commit on this branch has one, this one was just missed. Amended with `git commit --amend --signoff` to add it.

Verified locally: `pnpm build` (addon) and `pnpm lint` are both clean (0 errors). Local `test:ember` still has the known pre-existing local-runner flakiness, so relying on CI's `test` job for full suite confirmation.

patricklx:
Icons are not being shown in docs example rendering

patricklx:
Addressed in d7939112:

**Root cause of "Icons are not being shown in docs example rendering":** every lazily-loaded icon (`Folder`, `Document`, and in fact every other icon in the library — `Search`, `Task`, etc.) goes through `TrackedPromise.getValue()` (`src/utils/tracked.ts`), which calls `load()` (the dynamic `import()`) exactly once and caches the resulting promise forever. There was no `.catch()` — if that import ever rejects, the rejection was silently swallowed, `value` never gets set, and since `promise` stays set, no later `getValue()` call would ever retry. The icon then never renders, permanently, with nothing but an unhandled-rejection console warning to hint why. `Folder`/`Document` are the first *new* dynamic-import paths this PR introduces (unlike `Search`/`Task`, which are already exercised elsewhere in the docs and so are already "warm"), which is consistent with them being the ones that hit this.

**Fix:** handle the rejection branch — log the error and clear `promise` so the next `getValue()` (triggered by any subsequent re-render) retries the load instead of being stuck forever. The success path is unchanged.

**Also added:**
- A stress test in `tree-view-test.gts` mirroring the docs "With icons" example (mixed parent/leaf `Folder`/`Document` icons at multiple depths, some already expanded), asserting every one actually becomes *visible* (not just present in the DOM).
- A unit test for `TrackedPromise`'s new retry-after-rejection behavior.

**Verification:** `pnpm build` (types + rollup) and `pnpm lint` for the addon are both clean (0 errors). The local Playwright test runner (`test-app`'s `run-tests.mjs`) also ran cleanly this time end-to-end: 216/216 passing, including the 2 new tests. Local docs-app dev server/build still hang on this machine regardless of branch (pre-existing, unrelated to this change), so I couldn't re-screenshot the literal live-preview page directly — the fix instead targets the concrete, reproducible bug in the shared icon-loading utility that plausibly explains the symptom.

patricklx:
Icons are still not working. See docs-app/app/routes/application.ts. they need to be registered.
Update agent.md with that info.
Also,  initially Claude missed some apis and to mimic the story book from carbon react.  Check the script and check the prompt for Claude and improve it

patricklx:
Addressed the remaining feedback in 2d462f54:

**Icons are now actually fixed.** The previous fix (d7939112) handled the dynamic-import retry, but that was only half the problem. `docs-app/app/routes/application.ts` resolves `carbon-components-ember/icons` imports for live-preview docs examples against a **static, hand-maintained map** (kolay's `resolve` config) rather than doing real module resolution — `Folder`/`Document` were never added to that map at all, so no amount of import-retrying would have helped; kolay had nothing to resolve them to. Registered both in the import list and the `resolve` object, matching the existing `Bookmark`/`Task`/etc. entries.

**Updated `AGENTS.md`** with a new "Pitfall 4" explaining this: any new icon used in a `docs-app` live-preview example must be added to that static map in `docs-app/app/routes/application.ts` or it silently fails to render (no build/lint error, since neither checks this). Added it to the implementation checklist too.

**Improved the parity-automation prompt** (`.github/workflows/templates/agent-prompt.md`, plus the `fix-parity-issue.sh` fallback) per the feedback that I initially missed some APIs and didn't mirror the full Carbon React storybook:
- Phase 1 now requires explicitly enumerating every prop from the actual React `.tsx` source (not just what the default story exercises) and every story export in the `*.stories` file, before implementation starts — later phases (API comparison, docs) now check against that list instead of "match React" in the abstract.
- The docs step now explicitly calls for one live-preview example per enumerated story, not just a single default example.
- Fixed a stale/incorrect docs path in the template (`docs-app/app/routes/components/` doesn't exist; docs actually live at `docs-app/public/docs/2-components/*.md`).
- Added the icon-registration step to the documentation phase and success criteria.

**Verification:** `pnpm build:carbon` (types + rollup) succeeds, `pnpm lint` for the addon is clean (0 errors, pre-existing warnings only). `docs-app`'s `glint`/type-check has pre-existing errors (scss imports, `import.meta.hot`, `window.define`, etc.) that reproduce identically on a clean checkout with none of my changes applied — confirmed via `git stash` — so that's the known pre-existing local docs-app breakage, not something introduced here. The `application.ts` diff itself is a minimal, mechanical addition matching the existing pattern exactly (two new imports, two new object-literal entries).

patricklx:
ci is failing

patricklx:
Fixed in 9fae761e (already pushed, CI now green).

Root cause of "ci is failing": `.parity-check-data.json` had a missing opening quote on `Search`'s `lastCheckedCommit` key (introduced when #655 landed), making the file invalid JSON — this broke CI's required `test` check identically on `main`, not just this branch. Also fixed a separate flaky failure in "ProgressIndicator > should render invalid and disabled steps": it asserted on the lazily dynamic-imported Warning icon's class right after `settled()`, but dynamic `import()` isn't tracked by Ember's test waiters, so the icon wasn't guaranteed to exist yet — fixed with the same `waitUntil` pattern already used for lazy icons in `tree-view-test.gts`.

Current status: `test`, `ember-try` (beta/canary/release), `DCO`, and `Check Release Plan` are all passing.

PR #662, opened 2026-07-22
test: await settled() in OverflowMenu eventType regression test

patricklx:
## Summary
- Small follow-up to #656 / #469: the `eventType='mousedown'` regression test added while addressing review feedback dispatched the event and asserted immediately, without awaiting `settled()`. That's flaky since the dropdown's open-state update isn't guaranteed to be synchronous.

## Test plan
- [x] `pnpm build` passes


patricklx:
Rebase onto main

PR #668, opened 2026-07-24
feat: implement Popover component

patricklx:
## Summary

Implements the `Popover` and `PopoverContent` components (previously missing), closing #476.

- Controlled component: consumer owns `@open` and toggles it; `Popover` calls `@onRequestClose` on outside click or Escape (while focus is inside the popover content).
- Supports `@align` (all 20 alignment values from React, including deprecated `top-left`/`bottom-right`/etc aliases, mapped to the new `top-start`/`bottom-end`/etc values), `@caret`, `@dropShadow`, `@border`, `@highContrast`, `@backgroundToken`, `@isTabTip`, and `@as` (custom element type).
- `@autoAlign`/`@autoAlignBoundary` are implemented as a lightweight, dependency-free approximation of React's `floating-ui`-powered auto align: on open, it checks whether the popover content overflows the viewport (or a provided boundary element) and flips to the opposite side if so. This re-checks once per open rather than continuously tracking position via `floating-ui`, and picks from the same static alignment classes rather than computing arbitrary pixel offsets. `@alignmentAxisOffset` is accepted for API-signature parity but has no effect (documented as such), since it's an experimental floating-ui-only concept with no equivalent without that dependency.
- `PopoverContent` renders the `cds--popover`/`cds--popover-content`/`cds--popover-caret` structure matching Carbon's CSS positioning classes (verified against `@carbon/styles`'s `_popover.scss`, which is purely class-driven for the non-autoAlign case).

## Test plan

- [x] `cd carbon-components-ember && pnpm build` succeeds
- [x] `npx glint` passes for the new component (pre-existing repo-wide `moduleResolution` quirk affects test-app type-checking for all components equally, not specific to this change)
- [x] Added `test-app/tests/components/popover-test.gts` covering classes, alignment mapping, tab-tip trigger class, custom element type, outside-click close, inside-click no-close, and Escape-to-close
- [ ] Local Playwright test runner could not be run (`pnpm test`) — known pre-existing harness crash unrelated to this change
- [x] Docs added at `docs-app/public/docs/2-components/popover.md` with one live-preview example per Storybook story (Default, appearance options, TabTip, ExperimentalAutoAlign) plus API reference blocks
- [x] `Checkbox`/`Settings` icons used in docs examples registered in `docs-app/app/routes/application.ts`


patricklx:
Docs app preview is failing:
/ember-repl-c7bde0d2-9ad3-5b05-86af-8b56f3e2f5e2.js: Attempted to resolve a modifier in a strict mode template, but that value was not in scope: on: | | {{on 'click' (fn (set context 'open') (not context.open))}} | (error occurred in '/ember-repl-c7bde0d2-9ad3-5b05-86af-8b56f3e2f5e2.js' @ line 14 : column 8)

patricklx:
Fixed the docs preview failure: the four `gjs live preview` examples in `popover.md` used `{{on 'click' ...}}` on the trigger button without importing `on` from `@ember/modifier`. Strict-mode templates require modifiers to be explicitly imported, so the compiled preview couldn't resolve `on` in scope, causing the error.

Added `import { on } from '@ember/modifier';` to each of the four examples (Default, appearance options, TabTip, ExperimentalAutoAlign) in d1657b85.

patricklx:
Popover docs app preview fails with an error.
Undefined is not qn object. Probably missed component registration in docs app.
Portal docs page also doea not work. Fix it as well

patricklx:
Fixed both issues:

- **Portal docs preview blank/broken**: the live-preview sandbox (kolay) renders demos inside a shadow DOM. The Portal example defaulted to `document.body` as its target, so the portaled content escaped the shadow root and rendered unstyled at the very top of the real page — leaving the preview panel itself empty. Updated the example (`docs-app/public/docs/2-components/portal.md`) to capture a local element via the `did-insert` modifier and pass it as `@container`, so the demo now visibly renders inside the preview.
- **Popover preview "Undefined is not an object"**: this was caused by the same underlying gap — `did-insert` (from `@ember/render-modifiers`) wasn't resolvable inside the live-preview sandbox, so any snippet trying to use it (needed for the Portal fix above) threw at render time. Registered `didInsert` on the docs-app's `docs-support` resolve entry (`docs-app/app/routes/application.ts`), the same mechanism already used for `ThemeSupport`/`ThemeSwitcher`.

Verified by building docs-app locally with `DOCS_URL=pr-previews/pr-668` (matching the real preview build), serving it, and driving both pages with Playwright — no console/page errors on load or after interacting with every button/checkbox on the Popover page, and the Portal example now correctly shows the portaled content in its target box instead of a blank page.

Also re-ran `pnpm build:carbon`, `pnpm lint`, and `pnpm test:ember` (249/249 passing) to confirm no regressions.

patricklx:
Popover docs app demo not showing button for popover, cannot click on anything to toggle a popover

patricklx:
Found and fixed the root cause of the blank Popover demo (\"not showing button for popover, cannot click on anything to toggle a popover\").

**Root cause:** `pnpm create-autogenerated-files` (run by the PR-preview CI workflow before building docs-app, but not part of a normal `pnpm build`) regenerates `carbon-components-ember/src/components/index.ts` from a directory scan of `src/components/`. The generator only ever emitted the *default* export per file — so it silently rewrote the `Popover` line from:

```ts
export { default as Popover, PopoverContent } from './popover.gts'
```

to:

```ts
export { default as Popover } from './popover.gts'
```

dropping `PopoverContent` entirely. In the deployed PR-preview, that made `PopoverContent` resolve to `undefined` inside every live-preview sandbox on the Popover page, which crashed Glimmer's component-manager resolution (`Cannot convert undefined or null to object`) as soon as it tried to compile `<PopoverContent>` — leaving the whole preview area blank with no trigger button, on all four examples.

I confirmed this by running the exact CI sequence locally (`pnpm create-autogenerated-files` → build → serve → drive with Playwright): the crash reproduced 1:1, including the same error message and zero rendered buttons. Reverting to the fixed generator made it disappear.

**Fix (`carbon-components-ember/scripts/create-files.mjs`):** the generator now also scans each component file for additional `export class`/`export const`/`export function` declarations and includes them alongside the default export, so `PopoverContent` (and any future named exports) survive regeneration. This also fixes the same latent bug for `Layout`'s `LayoutConstraint` (previously only surviving by luck because nobody had re-run the generator since it was hand-added), and additionally now correctly captures `Icon`'s `registerIcon` and `ShapeIndicator`'s `ShapeIndicatorKinds`, which were exported from their source files but had been missing from the hand-maintained index all along.

Verified: `pnpm build:carbon`, `pnpm lint` (clean), `pnpm test:ember` (279/279 passing), and a local rebuild + Playwright check of the Popover docs page (button renders and toggles correctly in all 4 examples, zero console/page errors).

PR #670, opened 2026-07-25
feat: implement OrderedList component

### LINKED ISSUE #468: [Parity Check] Investigate OrderedList component

**Component**: OrderedList
**Commit**: 1110000

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/OrderedList
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-orderedlist--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/1110000e8374c35b7e040b09a7e292b2227f331d/packages/react/src/components/OrderedList)

---
*Auto-generated by parity-check script on 2026-07-14T19:57:12.660Z*

patricklx:
## Summary
- Adds `OrderedList`, matching Carbon React's `OrderedList` component (props: `nested`, `native`, `isExpressive`, plus pass-through HTML attributes/children — same shape as the existing `UnorderedList`).
- Renders `cds--list--ordered` (or `cds--list--ordered--native` when `@native` is set), with `cds--list--nested` and `cds--list--expressive` modifiers.
- Adds docs with a live-preview example for each React story (Default, Nested, NativeListStyles).
- Adds integration tests covering rendering, class toggling, and attribute pass-through, plus visual style snapshots (light/dark).

Closes #468

## Test plan
- [x] `pnpm build` succeeds
- [x] `pnpm test -- --filter="OrderedList"` — all 7 tests pass
- [x] Checked for `.gts` filename collisions — none introduced


patricklx:
Nested list is not indented. Native list style looks the same as first example

patricklx:
Thanks for the review! Both issues traced back to the same root cause, fixed in 379174e7:

- **Nested list not indented / native looking identical to the default:** the docs' live-preview blocks passed `@nested`/`@native` as bare shorthand (e.g. `<OrderedList @nested>`), which this app's live-preview compiler evaluates to an empty string rather than `true`. Since `""` is falsy, neither modifier class was ever actually applied in the rendered demos. Updated the doc examples to `@nested={{true}}` / `@native={{true}}`.
- **Default (non-native) list showed no numbers at all:** Carbon's stylesheet only renders the counter-based `::before` marker on `<li class="cds--list__item">` — plain `<li>` gets none. Since Ember can't clone yielded block content the way React clones `<ListItem>` children, `OrderedList` now applies `cds--list__item` to its direct `<li>` children itself via a modifier (backed by a `MutationObserver` so items added after the initial render are also picked up).

Verified with a real build of the docs app (not the dev server, which we've seen misbehave) — nested levels now indent and letter (a., b., …) as expected, and native vs. default are visually distinct. Added test coverage for the class being applied both initially and to dynamically-added items; updated the two style snapshots to reflect the now-correct computed styles. Full suite (274 tests) passes, `lint`/`glint` are clean.

PR #683, opened 2026-07-27
feat: implement NumberInput component

### LINKED ISSUE #630: [Parity Check] Investigate NumberInput component

**Component**: NumberInput
**Carbon React Version**: 11.112.0
**Commit**: 5102bd8

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/NumberInput
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-numberinput--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/5102bd860864c2580048183ffaf50d1ffd4400b5/packages/react/src/components/NumberInput)

---
*Auto-generated by parity-check script on 2026-07-15T17:51:03.906Z*

patricklx:
## Summary

- Implements `NumberInput` for parity with Carbon React: a number field with `+`/`-` stepper buttons, min/max/step clamping, and disabled/readOnly/invalid/warn states.
- Adds tests covering rendering, stepping (increment/decrement/step-size/clamping), controlled/uncontrolled value, disabled/readOnly, and validation states.
- Adds a docs page with a live preview and API reference.

## Notes / limitations

- Only the standard `type="number"` mode is implemented. React's experimental locale-aware `type="text"` mode (backed by `NumberFormatter`/`NumberParser` for BCP47 locale parsing/formatting) is not ported — it's a large, mostly orthogonal i18n feature and out of scope for keeping this component simple.
- Like React, the component becomes invalid automatically when the value is outside `min`/`max`, even without passing `@invalid` explicitly.

Closes #630

## Test plan

- [x] `pnpm build` succeeds
- [x] `pnpm test` — all 375 tests pass, including new NumberInput tests
- [x] Manually reviewed rendered markup against `@carbon/styles`' `_number-input.scss` selectors for class-name parity


patricklx:
Controlled input is not updating context.value in docs app preview

patricklx:
Thanks for flagging this @patricklx — dug into it.

**The component's controlled-value handling is correct.** I added a rendering test (c0f11ee) that mirrors the docs demo exactly — `@value` bound to a `TrackedObject` property, updated via `@onChange` — covering both stepper clicks and typed input, and it passes:

```
input value: 0 -> click up -> 1 -> type "42" -> 42 -> click up -> 43
```
with the external `context.value` staying in sync at every step. `internalValue` is only written when `@value` is `undefined` (`setValue` in `number-input.gts`), which is correct controlled-component behavior — the same as the other input components in this repo.

**What I think you actually saw:** I drove the live PR-preview deployment (https://ibm.github.io/carbon-components-ember/pr-previews/pr-683/) with Playwright to reproduce it directly, rather than relying on a manual check. The live-preview mount (`.glimdown-render`) intermittently renders completely empty — no inputs at all, just the syntax-highlighted source snippet below it — and this reproduces identically on other components' demos on the same preview (e.g. TextInput), not just NumberInput. I also saw a stray 404 for `https://ibm.github.io/assets/docs-app.css` (missing the PR-preview path prefix) while poking at it, which may be related. This looks like a pre-existing docs-app/kolay live-preview infra issue on the PR-preview deployment, not a regression from this PR.

Let me know if you're still seeing it after a hard refresh — happy to keep digging if there's actually something NumberInput-specific I'm missing.

patricklx:
accessing docs page for this:
main-CPNU7ZTl.js:121295 TypeError: Cannot read properties of undefined (reading 'TrackedMap')
    at 8006bc08-2877-4d08-aa1e-7365811e730d:9:23

patricklx:
Found the root cause of the `TrackedMap` crash: the docs demo imported `TrackedObject` from `tracked-built-ins`, which is only a **transitive** dependency here (not declared in `docs-app`), so it isn't resolvable inside the live-preview sandbox — hence `Cannot read properties of undefined (reading 'TrackedMap')`.

Every other component demo in this repo uses the project's own `trackedObject()` helper from `@ember/reactive/collections` for local reactive state instead. Switched the NumberInput demo to match that convention (bd375521):

```diff
-import { TrackedObject } from 'tracked-built-ins';
+import { trackedObject } from '@ember/reactive/collections';

-const context = new TrackedObject();
+const context = trackedObject();
```

This also explains the earlier "controlled input is not updating context.value" report — the component's controlled-value logic itself is correct (still covered by the regression test added earlier), but the demo's reactive context object wasn't reliably wired up in that sandbox.

Verified:
- `pnpm build` succeeds
- Full `test-app` suite passes (439/439), including all NumberInput tests
- Manually confirmed the docs-app package doesn't declare `tracked-built-ins` as a dependency — it was only ever available by transitive luck in other environments

Pushed to the branch.

PR #688, opened 2026-07-30
chore(docs-app): upgrade @universal-ember/docs-support to 0.7.0

patricklx:
## Summary
- Bumps `@universal-ember/docs-support` 0.6.7 → ^0.7.0, plus its own stated peer bumps (`ember-primitives` ^0.32.0 → ^0.46.0, `ember-mobile-menu` ^5.1.0 → ^6.0.0).
- Adds `@ember/test-waiters` to vite's `resolve.dedupe` in `docs-app/vite.config.mjs`, fixing a `Missing "./index.js" specifier in "@ember/test-waiters"` build error that surfaces once the dependency graph gets a second copy of that package after the bump.
- Verified locally (built + served `docs-app/dist`, screenshotted with Playwright) that the home page, a plain markdown page, and a component page with a live demo all render **identically** to the currently published site at https://ibm.github.io/carbon-components-ember/.

## Why it stops at 0.7.0 (not 0.9.x)
I initially tried to go all the way to the latest `docs-support` (0.9.12), per the task and referencing `ember-primitives` for guidance. That pulls in `kolay` ^5.1.2, which hard-depends on `ember-repl` ^8.0.2. That combination has several upstream bugs against this repo's current `@embroider/vite` + `ember-source` 6.4 stack:
- the published `ember-repl` 8.x npm packages are missing their `addon-main.cjs` / `ember-addon.main` entirely (embroider can't classic-resolve them),
- `ember-repl`'s "known modules" registry references `@ember/reactive` and other APIs that don't exist until later `ember-source`/`@embroider` releases than what's installed here,
- `repl-sdk` 1.6.1 (pulled in by `ember-repl` 8.x) calls `renderComponent` from `@ember/renderer`, which also isn't exported until a newer `ember-source`.

Getting 0.9.x working would require locally patching three separate npm packages (`ember-repl` twice, for two versions that both ended up in the tree, plus `kolay` for an unrelated manifest-group bug I hit along the way). That's too fragile/unmaintainable for this PR, so I stopped at 0.7.0 — the newest version that still uses the same Tailwind-based architecture and `kolay` 3.x line as before, and builds/renders cleanly with zero patches.

This also means `ember-repl` stays at `6.0.0` here rather than the `8.1.0` in dependabot PR #397 — that PR's target version is affected by the same issues above, so I'd recommend closing #397 rather than merging it as-is.

## Test plan
- [x] `pnpm build:carbon` — addon build unaffected
- [x] `pnpm build:docs` (with `DOCS_URL=versions/main`, matching CI) — builds cleanly, no warnings beyond pre-existing Sass deprecation noise
- [x] Served the built `docs-app/dist` locally and compared against https://ibm.github.io/carbon-components-ember/ with Playwright screenshots:
  - [x] Home page — pixel match
  - [x] `/1-get-started/index.md` (plain markdown page, sidebar nav) — pixel match
  - [x] `/2-components/accordion.md` (live component demo + source panel + API reference) — renders and is interactive, no console errors


patricklx:
Update: upgraded further, to `@universal-ember/docs-support@0.9.12` / `ember-repl@8.2.1` / `kolay@5.4.0` / `ember-primitives@0.61.1`.

**What changed since this stopped at 0.7.0:**

`ember-source` is now 6.12.0 (merged separately), which resolves two of the three blockers noted above — `@ember/reactive` and `renderComponent` from `@ember/renderer` both exist now.

The third blocker (`ember-repl` 8.x missing its v1 addon-main shim) is a genuine upstream regression: `ember-repl` 6.0.0 and 7.3.5 both ship `addon-main.cjs` + `"ember-addon": {"main": ...}`, but every published 8.x release (8.0.0 through 8.2.1) drops both, so embroider can't classic-resolve the addon at all (`ember-repl ... is missing its addon main file`). Fixed with a `pnpm patch` restoring the shim from 7.3.5 (`patches/ember-repl@8.2.1.patch`), the same shim ember-primitives and other v2 addons in this tree still ship.

That unblocked the rest of the upgrade, which surfaced one more thing: `@universal-ember/docs-support` >= 0.8 dropped its Tailwind-based CSS architecture in favor of shipping pre-built plain CSS (imported as a side effect of its components, e.g. `Shell` imports `./site-css/site.css` etc. directly). That meant:
- `docs-app/config/tailwind.config.mjs` no longer delegates to the now-removed `@universal-ember/docs-support/tailwind` export; it's a small standalone config scoped to docs-app's own templates.
- `docs-app/config/postcss.config.mjs` now scopes the Tailwind PostCSS plugin to non-`node_modules` files — Tailwind's plugin errors on any bare `@layer` it finds in a file that doesn't also declare `@tailwind`, and docs-support's CSS uses `@layer` for plain cascade ordering, not Tailwind's directive.
- `app/styles/app.css` now declares `@tailwind base/components/utilities` itself and is explicitly imported from `app.ts` — previously that only worked because it was reached transitively through docs-support's own (now-removed) Tailwind-based `site.css`, and embroider/vite doesn't auto-pick-up the classic `app/styles/app.css` convention the way classic ember-cli did.

Verified with a clean install that both `pnpm build:carbon` and `DOCS_URL=versions/main pnpm build:docs` succeed, and that the built CSS contains both docs-app's own Tailwind utility classes and docs-support's site CSS. Home page renders correctly when served locally (screenshot-checked).

**Still to do before merge:** a full pixel-diff pass across component doc pages (especially ones using kolay's live REPL/playground demos, to confirm `ember-repl` 8.2.1 actually works end-to-end and not just that the build no longer errors) — I only spot-checked the home page this round. Local static serving of `docs-app/dist` also hit a pre-existing, unrelated crash (`window.define is not a function`) that I confirmed reproduces identically on the current unmodified `main`-based build too, so it's not a regression from this change, just a limitation of the simple local server I used for the check.

PR #767, opened 2026-08-12
feat: bring Tag component to parity with React

### LINKED ISSUE #718: [Parity Check] Update Tag component
## Component Update Required

**Component**: Tag
**Latest Commit**: a57cf8a

### Recent Changes in React

- [`93b1978`](https://github.com/carbon-design-system/carbon/commit/93b1978f4d7a07e159c02a938acf6ae8a0e5e985) refactor: remove stale no-console eslint links (#22205)

### Investigation Tasks

- [ ] Update tests
- [ ] Update documentation

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/a57cf8a896fe3fe09c5ab27648cd58947f37360a/packages/react/src/components/Tag)
- [Component Commits](https://github.com/carbon-design-system/carbon/commits/main/packages/react/src/components/Tag)

---
*Auto-generated by parity-check script on 2026-08-01T12:38:03.529Z*

patricklx:
Closes #718

## Summary
- Added `size` (`sm`/`md`/`lg`), `disabled`, `id`, `renderIcon`, `decorator`/`slug` (deprecated alias) args to `Tag`, matching React's `TagBaseProps`.
- Added the missing `high-contrast` and `outline` values to `type`.
- `className`/attribute passthrough (React's `className` prop) is already covered by `Element: HTMLDivElement` + `...attributes` — Ember's splattributes merge a passed `class` with the component's own `cds--tag*` classes automatically, no extra arg needed.
- Added docs examples for sizes, disabled state, icon and decorator.
- Added tests covering the new args.

## Out of scope
- `filter` / `onClose` / `title` are deprecated in React in favor of a separate `DismissibleTag` component, which doesn't exist in this addon yet and isn't tracked by the parity checker as its own component — left out of this PR rather than building deprecated behavior.
- React's `as` prop (polymorphic element, defaults to `button` when `onClick`/the operational class is present) isn't part of the public `TagBaseProps` interface — it only matters once an interactive/clickable Tag variant exists, which none of the React stories (`ReadOnly`, `Skeleton`, `withAILabel`) demonstrate and this Tag doesn't support an `onClick` arg. Left out; worth adding alongside `element` (from `ember-element-helper`) if/when an interactive Tag is implemented.
- React's automatic ellipsis-overflow tooltip (`isEllipsisActive` + `DefinitionTooltip`) isn't demonstrated by any story either. It could be built with a functional `modifier()` (see `-private/tooltip.gts`) for DOM measurement, so it's not that this repo lacks the pattern — it's left out purely to keep this change scoped to what the stories cover.

### DIFF
--- a/.parity-check-data.json
+++ b/.parity-check-data.json
@@ -531,7 +531,7 @@
     },
     "Tag": {
       "lastCheckedCommit": "5102bd860864c2580048183ffaf50d1ffd4400b5",
-      "lastSyncedCommit": "N/A",
+      "lastSyncedCommit": "a57cf8a896fe3fe09c5ab27648cd58947f37360a",
       "hasChanges": false,
       "changeCount": 0,
       "lastUpdate": null
--- a/carbon-components-ember/src/components/tag.gts
+++ b/carbon-components-ember/src/components/tag.gts
@@ -1,6 +1,36 @@
 import Component from '@glimmer/component';
+import { guidFor } from '@ember/object/internals';
+import type { ComponentLike } from '@glint/template';
 
 export type Args = {
+  /**
+   * Specify the id for the tag.
+   */
+  id?: string;
+  /**
+   * Specify if the `Tag` is disabled.
+   */
+  disabled?: boolean;
+  /**
+   * A component used to render an icon.
+   */
+  renderIcon?: ComponentLike;
+  /**
+   * Specify the size of the Tag. Currently supports either `sm`, `md`
+   * (default) or `lg` sizes.
+   */
+  size?: 'sm' | 'md' | 'lg';
+  /**
+   * **Experimental:** Provide a `decorator` component (e.g. AILabel) to be
+   * rendered inside the Tag.
+   */
+  decorator?: ComponentLike;
+  /**
+   * @deprecated please use `decorator` instead.
+   * **Experimental:** Provide a Slug/AILabel component to be rendered
+   * inside the Tag.
+   */
+  slug?: ComponentLike;
   type:
     | 'red'
     | 'magenta'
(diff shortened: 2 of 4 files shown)


patricklx:
icons are not correctly positions

patricklx:
Fixed the icon positioning issue.

**Root cause:** the addon's `Icon` base component defaults to a 24px SVG variant and applies a 5px margin unless `@size`/`@svgClass` are explicitly passed. `Tag`'s `<@renderIcon />` was invoked bare, so it rendered a 24px icon (plus margin) inside the 16px `.cds--tag__custom-icon` box, causing it to overflow/misalign. This matches the existing convention elsewhere in the addon (`number-input`, `notification`, `tree-view`, etc.), which always pass `@size='16'` + a `@svgClass` override for fixed 16px icon slots.

**Changes in `carbon-components-ember/src/components/tag.gts`:**
- `<@renderIcon @size='16' @svgClass='cds--tag__custom-icon-svg' />` — sizes the icon correctly and strips the base component's default margin (the override class is inert, chosen so it doesn't collide with `.cds--tag__custom-icon`'s own CSS, which would have reintroduced the same bug via `margin-inline-end`).
- Narrowed `renderIcon`'s type from a bare `ComponentLike` to `typeof Icon` (matching `tree-view.gts`'s `icon` arg), since invoking it with `@size`/`@svgClass` requires that signature.
- The docs "With decorator" example passed the same icon as both `@renderIcon` and `@decorator`, so it had the identical oversized-icon bug in the decorator slot. Fixed via `@decorator={{component Add size='16'}}` to curry the size in.

**Tests:** added a rendering test that loads real Carbon CSS and asserts the icon's `getBoundingClientRect()` fits inside its wrapper with no stray margin — verified it fails without the fix and passes with it. Full suite (469 tests) and lint pass.

PR #770, opened 2026-08-13
feat: implement SidePanel component

patricklx:
## Summary

- Implements `SidePanel`, previously missing from the Ember addon (`carbon-components-ember/src/components/side-panel.gts`), covering the full React prop surface: `actions`, `actionToolbarButtons` (leading/icon-only variants), `currentStep`/`onNavigationBack`, `includeOverlay`/`preventCloseOnClickOutside`, `slideIn`/`selectorPageContent`, `resizable`, `decorator` (plus deprecated `slug`/`aiLabel` handled the same way React treats them — as icon aliases, via the `icon`/`renderIcon` alias on toolbar buttons), `selectorPrimaryFocus`, `launcherButtonRef` (as a plain `HTMLElement` rather than React's `RefObject`, since Ember has no ref objects), `animateTitle`, and `size`/`placement`.
- Translates React's hooks/refs into this addon's existing idioms: functional `ember-modifier`s for DOM work (presence/exit-animation tracking, focus-on-open, Escape-to-close, resize drag/keyboard handling, and the scroll-driven title collapse), a `ComponentLike` arg for `@decorator`/toolbar icons, and a `subtitle` block for rich content alongside the plain `@subtitle` string arg.
- The scroll-driven title-collapse effect is implemented by setting the same CSS custom property (`--cds--side-panel--scroll-animation-progress`) Carbon's own stylesheet already keys off, rather than re-implementing React's per-element margin bookkeeping.
- Bumps `@carbon/styles` from `1.112.0` to `1.113.0` (already permitted by the existing `^1.112.0` range) — `1.113.0` is the first release whose compiled CSS actually includes `.cds--side-panel`/`.cds--resizer` styles; without it the component would render unstyled.
- `ActionSet`/`ButtonSet`/`Resizer`/`IconButton` are not implemented as separate public components here (they're tracked separately, e.g. `ButtonSet`/`IconButton` are already open parity issues) — the footer action-button stacking/ordering and the resize handle are implemented privately inside `side-panel.gts` instead, matching their real Carbon CSS classes.
- Adds docs at `docs-app/app/templates/2-components/side-panel.md` with one live-preview example per React Storybook story (default/slide-over, slide-in, action toolbar, multi-step, decorator, initial-focus selector, static title, no-title, plus resizable).
- Adds tests at `test-app/tests/components/side-panel-test.gts` covering rendering, open/close, actions, toolbar buttons, decorator, overlay/outside-click, Escape, slide-in, and resizing.

[...]

PR #791, opened 2026-08-20
feat: implement Resizer component

### LINKED ISSUE #747: [Parity Check] Investigate Resizer component

**Component**: Resizer
**Commit**: a57cf8a

### Investigation Tasks

- [ ] Review React implementation: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/Resizer
- [ ] Check Storybook examples: https://react.carbondesignsystem.com/?path=/docs/components-resizer--overview

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/a57cf8a896fe3fe09c5ab27648cd58947f37360a/packages/react/src/components/Resizer)

---
*Auto-generated by parity-check script on 2026-08-07T15:48:33.362Z*

patricklx:
## Summary

Implements `Resizer` for parity with Carbon React's [`Utilities/Resizer`](https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/Resizer) ([Storybook](https://react.carbondesignsystem.com/?path=/docs/components-resizer--overview)). This component didn't previously exist in `carbon-components-ember` (Path B).

- A draggable/keyboard-operable `<div role="separator">` handle. By default it resizes its previous/next DOM siblings directly (matching React's behavior); passing `@onResize` opts into fully-controlled mode instead (e.g. to drive a CSS Grid's `grid-template-columns`).
- Full React prop surface: `@orientation` (required), `@onResize`, `@onResizeEnd`, `@onDoubleClick`, `@thickness` (default 4), plus a default block for custom drag-handle content (`children` in React). `className`/native attribute spreading is handled by Ember's `...attributes` rather than a separate `@class` arg.
- `@onResizeEnd` receives the resizer's element directly rather than a `RefObject` — the natural Ember translation, since there's no `forwardRef` equivalent needed here.
- Idiom mapping (per AGENTS.md): the `useRef`/`useEffect` DOM wiring (sibling measurement, mouse drag, keyboard nav, double-click reset) is a single functional `ember-modifier`; the key-driven debounced `onResizeEnd` uses a restartable `ember-concurrency` task + `timeout()` instead of React's hand-rolled `debounce()`.
- Docs at `docs-app/app/templates/2-components/resizer.md` cover all 8 Storybook stories (single panel unbounded/bounded/overlay, two panels horizontal/vertical, four panels, fully-controlled CSS Grid, custom drag handle).
- Tests at `test-app/tests/components/resizer-test.gts` cover rendering/classes/thickness, default drag-resize behavior, controlled mode via `@onResize`, `@onResizeEnd` on mouse-up and (debounced) on keyboard resize, arrow-key step sizes, and double-click reset with/without `@onDoubleClick`.

## Note on parity tracking

`node scripts/parity-check.mjs --mark-synced Resizer` is a no-op: since Resizer is a brand-new component, it has no existing entry in `.parity-check-data.json`'s `componentMetadata` yet (same situation as a previous Menu PR). Populating it requires a full rescan, which is out of scope for this PR — the scheduled `parity-check-weekly.yml` workflow run will pick it up.

## Test plan

- [x] `cd carbon-components-ember && pnpm build` succeeds
- [x] `cd test-app && pnpm test` — full suite passes (546/546), including all new Resizer tests
- [x] No `.gts` filename collisions with other exported components

Closes #747


patricklx:
Confirmed the root cause: \`.cds--resizer\` has zero CSS rules in \`@carbon/styles\` 1.112.0 (the version this repo currently pins) — no background-color, cursor, or hover/focus/active states, so the handle renders but is visually invisible. \`@carbon/styles\` 1.113.0 is the first release that ships \`scss/components/resizer\`; bumped the pin to \`1.113.0\` in all three workspace \`package.json\`s (and regenerated the lockfile pinned to exactly 1.113.0, not the caret range's newest match, to stay clear of CI's pnpm \`minimumReleaseAge\` gate on same-day releases).

Verified the prebuilt \`@carbon/styles/css/styles.css\` (the bundle this repo actually loads at runtime) now contains the \`.cds--resizer\` rules and the \`--cds-border-subtle\`/\`--cds-border-interactive\` custom properties they depend on. Also checked every demo in \`resizer.md\` and confirmed each panel container is `display: flex` with a cross-axis that gives the handle a non-zero size in the orientation it uses.

Note: PR #770 (SidePanel) makes this exact same three-`package.json` + lockfile bump, since 1.113.0 also shipped `.cds--side-panel`. Whichever of #770/#791 merges second will hit a trivial lockfile conflict on the `@carbon/styles` bump — expected, not a real conflict.

(Unrelated, pre-existing, repo-wide: `pnpm build` currently fails locally on `src/icons.ts`/`icon-indicator.gts` with missing `@carbon/icons` module files — confirmed this also fails identically on `origin/main`, so it's not caused by or specific to this PR. Verified the CSS fix via the resolved `@carbon/styles` package contents directly instead.)

patricklx:
## Follow-up: "drag handle still not visible"

Investigated with a real local `DOCS_URL=versions/main pnpm build` + Playwright (screenshots + computed-style checks inside the demo's shadow root, not just "build passes"): the handle **does** render correctly — non-zero size, correct `border-subtle` background (confirms the earlier 1.113.0 CSS bump was the real fix for the previously-missing rules).

What's actually going on: `.cds--resizer` is a bare 4px `border-subtle`-colored bar with no other visual affordance, and that's not a gap in this port — it's upstream Carbon React's Resizer, byte-for-byte (same class names, same token, same lack of extra DOM). Their own MDX says so explicitly: *"The handle trigger cue can be enhanced if needed by writing a pseudo-class to make the trigger easier to grab, or any custom elements can be passed as children."* Every React story repeats a `::before` hit-target-widening rule for exactly this reason.

Fix (docs-only, no component/SCSS changes — that would diverge from upstream): ported that same `::before` pattern into all 8 demos on `resizer.md`, and added a short paragraph documenting the intentional subtlety + pointing at the existing "With a custom drag handle" demo. Rebuilt and re-verified in a real browser: all `::before` pseudo-elements now compute a real (20px) hit box instead of collapsing to 0 (worth noting: `var(--cds-spacing-*)` custom properties don't actually exist in `@carbon/styles`' shipped CSS the way React's Storybook docs environment apparently provides them — used literal rem values instead, matching `@carbon/layout`'s real spacing scale).

PR #792, opened 2026-08-20
fix: align Loading component with Carbon React API

### LINKED ISSUE #783: [Parity Check] Update Loading component
## Component Update Required

**Component**: Loading
**Carbon React Version**: 11.114.0
**Latest Commit**: 188d232

### Recent Changes in React

- [`485de61`](https://github.com/carbon-design-system/carbon/commit/485de61098e644bc3430c681de16f1a4872a2c80) fix(Loading): add Storybook controls (#22862)

### Investigation Tasks

- [ ] Update tests
- [ ] Update documentation

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/188d23202ec1092322dee92cf0df9d9958224ae4/packages/react/src/components/Loading)
- [Component Commits](https://github.com/carbon-design-system/carbon/commits/main/packages/react/src/components/Loading)

---
*Auto-generated by parity-check script on 2026-08-17T09:33:54.606Z*

patricklx:
Closes #783

## What changed

Investigated the full React `Loading` API (`Loading.tsx` + `Loading.stories.js`) and found the Ember component had drifted well beyond the single "add Storybook controls" commit the issue flagged:

- **`title` → `description`** — the old `title` arg was typed as `boolean` but used as a string; renamed/retyped to `description` matching React's `description` prop, default `'loading'`, applied to the svg `<title>` and `aria-label`.
- **`overlay` → `withOverlay`**, default flipped `false` → `true` to match `carbon-react`'s actual component default (`withOverlay = true`). The React Storybook `Default` story overrides this to `false` just for display purposes — the component's real default is `true`. **This is a behavior change**: a bare `<Loading />` now renders as a full-viewport overlay by default instead of a small inline spinner. Consumers relying on the old bare default need `@withOverlay={{false}}`.
- **`@active={{false}}` no longer removes the component from the DOM.** React keeps it mounted and switches to a "stopped" CSS state (`cds--loading--stop` / `cds--loading-overlay--stop`) with `aria-live="off"`; the overlay variant's own CSS (`display: none`) is what visually hides it, not a template `{{#if}}`.
- Added `aria-atomic`, `aria-live`, `role="img"`, `aria-label` to match React's accessibility markup, and updated the svg `viewBox`/circle geometry to React's current `0 0 100 100` / `r=42|44` values (small-variant circles were previously oversized relative to the small container).

`inline` and `classNames` are pre-existing extensions with no equivalent in `carbon-react`'s `Loading` (`InlineLoading` is tracked separately in `.parity-check-data.json` as a missing component, not covered by this issue). Their behavior — including fully hiding when `@active` is false — is preserved unchanged, since `Button`, `Icon`, and `DataTable`'s search input all depend on that exact behavior internally.

## Test plan
- [x] `pnpm build` succeeds (types + rollup)
- [x] `glint` reports no errors for `loading.gts`
- [x] Full `test-app` suite: 543/543 passing, including `Button`, `IconIndicator`, and `DataTable` tests that render `Loading` internally
- [x] New `loading-test.gts` covers: default (no overlay), `@active=false` (stop, not hide), `@small`, `@description`, `@withOverlay` default-true, `@withOverlay` + `@active=false` (hidden via CSS), `@inline`, and `@inline` with no `@description` (verifies the default description text does not leak into the visible inline text node)
- [x] Docs live-preview example updated to the new arg names and covers every combination from the single React `Default` story (active, small, description, withOverlay)

### DIFF
--- a/.parity-check-data.json
+++ b/.parity-check-data.json
@@ -426,7 +426,7 @@
     },
     "Loading": {
       "lastCheckedCommit": "5102bd860864c2580048183ffaf50d1ffd4400b5",
-      "lastSyncedCommit": "N/A",
+      "lastSyncedCommit": "188d23202ec1092322dee92cf0df9d9958224ae4",
       "hasChanges": false,
       "changeCount": 0,
       "lastUpdate": null
--- a/carbon-components-ember/src/components/loading.gts
+++ b/carbon-components-ember/src/components/loading.gts
@@ -4,8 +4,8 @@ import { defaultArgs } from '../utils/decorators.ts';
 export type Args = {
   active?: boolean;
   small?: boolean;
-  overlay?: boolean;
-  title?: boolean;
+  withOverlay?: boolean;
+  description?: string;
   inline?: boolean;
   classNames?: string;
 };
@@ -19,8 +19,8 @@ export default class LoadingComponent extends Component<LoadingComponentSignatur
   args: Args = defaultArgs(this, {
     active: true,
     small: false,
-    overlay: false,
-    title: undefined,
+    withOverlay: true,
+    description: 'loading',
     inline: false,
     classNames: '',
   });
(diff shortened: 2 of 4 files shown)


patricklx:
Inline loader is not inline

patricklx:
Fixed in 3b773f60.

`@inline`'s wrapper had a hardcoded `display: inline-block` inline style, which (by CSS specificity) always overrides the real `.cds--inline-loading` class's `display: flex`. Since the animation and text are separate block-level `<div>`s, they were stacking vertically instead of sitting side by side — hence "not inline".

Changed the wrapper's inline style to `display: inline-flex; align-items: center`, keeping the existing `width: initial; margin-left: 1rem` shrink-to-fit sizing that Button/Icon/DataTable's search input rely on internally, while making the icon and description text actually lay out on one line.

Added a regression test asserting the wrapper's computed `display` is `inline-flex`. Full `test-app` suite still passes 543/543 (including Button, Icon, and DataTable's internal `Loading` usages), and `glint`/`pnpm build` are clean.

PR #793, opened 2026-08-21
feat: align CopyButton with React's public API

### LINKED ISSUE #751: [Parity Check] Update CopyButton component
## Component Update Required

**Component**: CopyButton
**Latest Commit**: a57cf8a

### Recent Changes in React

- [`b94c6cf`](https://github.com/carbon-design-system/carbon/commit/b94c6cf69028ed6e107cf36a8638546f87f2fef0) chore(CopyButton): add Storybook controls (#22805)

### Investigation Tasks

- [ ] Update tests
- [ ] Update documentation

### Resources

- [Carbon React Component](https://github.com/carbon-design-system/carbon/tree/a57cf8a896fe3fe09c5ab27648cd58947f37360a/packages/react/src/components/CopyButton)
- [Component Commits](https://github.com/carbon-design-system/carbon/commits/main/packages/react/src/components/CopyButton)

---
*Auto-generated by parity-check script on 2026-08-07T15:48:44.094Z*

patricklx:
Closes #751

## What changed

`CopyButton` only exposed `targetElementId`/`targetElement`/`inline` and hardcoded its tooltip text, `feedbackTimeout` (3000ms instead of React's 2000ms) and `iconDescription` default ("Copy code" instead of React's "Copy to clipboard"). It was missing `align`, `autoAlign`, `disabled`, `feedback`, `feedbackTimeout`, `iconDescription` and `onClick`, all present on React's `CopyButton`.

- Added the missing args, matching React's `CopyButton.tsx`/`Copy.tsx` defaults.
- Rebuilt the tooltip on the addon's own `Popover`/`PopoverContent` (`components/popover.gts`), which already implements Carbon's full alignment set including the deprecated→new alignment mapping, instead of the private floating-ui-based `-private/tooltip.gts`. This is the pattern already used by `IconIndicator`'s compact mode.
- Replaced the bare `setTimeout` feedback-hide timer with a restartable `ember-concurrency` task, and the `did-insert` element capture with a functional `ember-modifier`, per AGENTS.md.
- `CodeSnippet` previously wrapped `<CopyButton>` in a hand-rolled `<span class="cds--popover-container ...">` — removed, since `CopyButton` now renders its own popover container with the same classes.
- Kept `targetElementId`/`targetElement`/`inline` as Ember-side conveniences (React's `CopyButton` doesn't do the actual clipboard write itself — that's left to the consumer's `onClick` — but this addon's existing `CopyButton` already performs the copy, and is consumed internally by `CodeSnippet`, so removing that would be a breaking change with no React equivalent to replace it).

## Testing

- `pnpm build` passes.
- Added `test-app/tests/components/copy-button-test.gts` covering the default render, copy + feedback tooltip, custom `feedback`/`iconDescription`, `onClick`, `disabled`, `align`, `targetElementId`, and `inline`. Full suite (543 tests) passes.
- Added `docs-app/app/templates/2-components/copy-button.gjs.md` with live-preview examples (default, custom feedback, alignment/disabled).

### DIFF
--- a/.parity-check-data.json
+++ b/.parity-check-data.json
@@ -377,7 +377,7 @@
     },
     "CopyButton": {
       "lastCheckedCommit": "5102bd860864c2580048183ffaf50d1ffd4400b5",
-      "lastSyncedCommit": "N/A",
+      "lastSyncedCommit": "188d23202ec1092322dee92cf0df9d9958224ae4",
       "hasChanges": false,
       "changeCount": 0,
       "lastUpdate": null
--- a/carbon-components-ember/src/components/code-snippet.gts
+++ b/carbon-components-ember/src/components/code-snippet.gts
@@ -55,11 +55,7 @@ export default class CarbonCodeSnippet extends Component<CarbonCodeSnippetSignat
             {{~yield~}}
           </PreCode>
         </div>
-        <span
-          class='cds--popover-container cds--popover--caret cds--popover--high-contrast cds--popover--bottom cds--tooltip cds--icon-tooltip'
-        >
-          <CopyButton @targetElement={{this.carbonElement}} />
-        </span>
+        <CopyButton @targetElement={{this.carbonElement}} />
       </div>
     {{/if}}
     {{#if (eq @type 'multiline')}}
@@ -83,11 +79,7 @@ export default class CarbonCodeSnippet extends Component<CarbonCodeSnippetSignat
           </PreCode>
         </div>
         <div class='cds--snippet__overflow-indicator--right'></div>
-        <span
-          class='cds--popover-container cds--popover--caret cds--popover--high-contrast cds--popover--bottom cds--tooltip cds--icon-tooltip'
-        >
-          <CopyButton @targetElement={{this.codeElement}} />
-        </span>
+        <CopyButton @targetElement={{this.codeElement}} />
         <button
           {{on 'click' (fn (set this 'expanded') (not this.expanded))}}
           class='cds--btn cds--btn--ghost cds--btn--sm cds--snippet-btn--expand'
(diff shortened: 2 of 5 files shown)


patricklx:
The icons in copy button does look strange. Maybe wrong size or wrong alignment

patricklx:
Fixed the icon rendering issue @patricklx flagged: `f00851be`

**Root cause:** when a block is passed without `@inline` (the default icon-only mode — which is exactly what every live-preview demo on the docs page does, e.g. `<CopyButton>console.log('hello world');</CopyButton>`), the component rendered the `<code>{{yield}}}</code>` block *visibly* alongside the SVG icon inside the same flex-centered 40x40 button. That pushed the icon off-center (icon was offset ~8px right instead of centered) and made it look wrong/misaligned — exactly what you saw on the preview.

**Fix:** the block's yielded text is only meant to supply the copy source in icon-only mode (it's the Ember-side convenience mentioned in the docs — "pass the text to copy as the block content"), not to be displayed. It's now hidden with the addon's existing `cds--visually-hidden` utility (same pattern used by `IconIndicator`'s compact-mode label) whenever `@inline` isn't set, so the text stays available for `copyToClipboard()`'s `.textContent` read but no longer shows up next to the icon. `@inline` mode (used by `CodeSnippet`'s inline variant) is untouched — the code text still renders visibly there, as intended.

Verified by building the addon + docs-app and screenshotting the live-preview demo in a real browser (before/after): the icon is now centered at (12,12) in the 40x40 button instead of (20,12). Added a regression test (`copy-button-test.gts`) asserting the block gets `cds--visually-hidden` when not inline. Full suite (544 tests) and glint pass.

PR #835, opened 2026-09-09
feat: implement DatePicker and DatePickerInput components

patricklx:
## Summary

- Implements `DatePicker` and `DatePickerInput`, closing #426 and #427. Split off from #641 (which covered Dropdown/FileUploader/Layer) since wrapping [flatpickr](https://flatpickr.js.org/) is a real new-runtime-dependency decision that deserved its own review.
- `DatePicker` supports React's three `datePickerType`s: `simple` (plain text field, no calendar), `single`, and `range` (two fields sharing one calendar via flatpickr's `rangePlugin`). Also implements `dateFormat`, `minDate`/`maxDate`, `allowInput`, `closeOnSelect`, `readOnly`, `short`, deprecated `light`, and `onChange`/`onOpen`/`onClose`.
- `DatePickerInput` is the labelled text field `DatePicker` composes and yields (pre-bound to `datePickerType`/`readOnly`) — one for `simple`/`single`, two for `range`. Supports `labelText`/`hideLabel`, `invalid`/`invalidText`, `warn`/`warnText`, `helperText`, `size`, `disabled`, and a `decorator` `ComponentLike` slot.

### Architecture decision: flatpickr wiring

- Added `flatpickr` (`^4.6.13`) as a direct dependency — confirmed nothing in the monorepo already pulls it in transitively, and no existing Ember flatpickr wrapper addon covers Carbon's specific calendar markup/class requirements (`cds--date-picker__*` classes layered onto flatpickr's own DOM, per Carbon React's `updateClassNames`).
- Wired in as a real `ember-modifier` (`attachFlatpickr`) that constructs the flatpickr instance once the input(s) are in the DOM and destroys it on teardown or when structural config changes — matching AGENTS.md's "Use Modifiers for DOM Work" pattern rather than `did-insert`/`did-update`. The component never touches `document`/`window` outside that modifier, so it's inert during SSR/fastboot.
- A separate `syncValue` modifier applies a controlled `@value` onto the *existing* flatpickr instance (via `setDate`) without tearing it down/rebuilding — otherwise every parent re-render from a consumer's own `onChange` would kill the calendar mid-interaction. Covered by a dedicated test (`the calendar stays interactive across two selections when @value is controlled`).

## Test plan

- [x] `pnpm build` in `carbon-components-ember/` (includes `glint --declaration`) — passes
- [x] `pnpm glint` — passes
- [x] `pnpm lint` — passes (only pre-existing, unrelated warnings)
- [x] `pnpm build` in `test-app/` — passes
- [x] `pnpm test` in `test-app/` — 567/567 passing, including 12 new DatePicker/DatePickerInput integration tests covering simple/single/range modes, controlled vs. uncontrolled `@value`, `minDate`, `readOnly`, flatpickr teardown, invalid/warn/helper text, and the decorator slot
- [x] Filename collision check (`find src/components -name "*.gts" ... | sort | uniq -d`) — no collisions
- [x] Did **not** run `parity-check.mjs --mark-synced` (no-ops for brand-new components not yet in `.parity-check-data.json`; next scheduled run picks these up)

FluidDatePicker/FluidDatePickerInput (#437/#438) intentionally left out of scope — noted as optional stretch work in the originating task.


patricklx:
Fixed both:

- **Arrows too big**: flatpickr's stock prev/next month arrows are a raw `viewBox`-only SVG with no `width`/`height`, and `@carbon/styles` only sizes the containing 40px button, not the svg - so they rendered at the browser's default replaced-element size. Swapped them for Carbon's own 16px `chevron-left`/`chevron-right` icons (matches Carbon React's behavior).
- **Shadow root**: flatpickr appends its calendar to `document.body` by default, unconditionally - it was never configurable here. Added an `@appendTo` arg (passed straight through to flatpickr's own `appendTo` option) so it can be redirected into your own shadow root, same pattern `Menu`'s `@target` already uses for its portal.

Added two regression tests covering both. Full test-app suite (575 tests) green.

patricklx:
Addressed: flatpickr's calendar now stays inside a shadow root by default, without requiring the consumer to pass `@appendTo`.

Previously the calendar only stayed in-shadow if you explicitly pointed `@appendTo` at an element inside that root — flatpickr's own default is `document.body`, which escapes any shadow root the picker itself is rendered in. `attachFlatpickr` now defaults the append target to the picker's own `.cds--form-item` wrapper (an explicit `@appendTo` still overrides this), so the calendar always lands in the same root the picker renders into.

Simplified the docs demo to drop the `context.container`/`didInsert` wiring that only existed to thread a shadow-root-local element into `@appendTo`.

Verified end-to-end with a real `DOCS_URL=versions/main` build served locally and driven with Playwright: the calendar renders inside `<carbon-shadow-demo>`'s shadow root, never escapes to `document.body`, and stays fully interactive — clicking a day selects the date and closes the calendar, with no page errors. Also added a regression test asserting the default append target and full `test-app` suite (579 tests) green.

PR #838, opened 2026-09-09
feat: port Carbon AI Chat Launcher and ChatShell components

patricklx:
## Summary
- First vertical slice of the Carbon AI Chat port (see `.pjp-runner/todo.md`'s todo #642 and #640): ports `@carbon/ai-chat-components`' `launcher` and `chat-shell` Lit widgets as Ember components (`Launcher`, `ChatShell`), matching their real public prop/attribute/event/slot surface per the published `custom-elements.json` manifest.
- Deliberately does **not** port the internal Lit managers (`ResizeObserverManager`, `CornerManager`, `PanelManager`, `WorkspaceManager`, `AriaAnnouncerManager`) — only the public surface and the resulting DOM/class structure. See AGENTS.md's new "Porting Carbon AI Chat" section for the full list of what's ported vs. deferred, and why (decisions on file location, filename-collision check, styling, and controlled/uncontrolled state are documented there).
- Adds SCSS for both components under `carbon-components-ember/src/styles/ai-chat/`, adapted from upstream's real (non-minified) `.scss` source fetched from GitHub, using `@carbon/styles`' compiled `--cds-*` custom properties for theming.
- Flips `createIssues: true` for the `carbon-ai-chat` source in `scripts/parity-check.mjs` now that a normal missing/implemented split makes sense (confirmed the `.parity-check-data.json` "already has a stored version" trap flagged in the task doesn't apply yet — this source has never persisted data before, so the next run will take the `missing` branch and file issues for all ~18 still-unported components).

## Test plan
- [x] `pnpm glint` (addon) — clean
- [x] `pnpm build:js` (addon) — clean
- [x] `pnpm lint` (addon: hbs/types/js) — clean (0 errors)
- [x] `ember-template-lint` on the two new test files — clean
- [x] Full `test-app` suite (Playwright/QUnit) — 573/573 passing, including the 17 new `ai-chat/Launcher` and `ai-chat/ChatShell` tests
- [x] Manually compiled `carbon-components-ember/src/styles/index.scss` with `sass` to confirm the new partials compile and emit the expected `.cds-aichat-*` rules


patricklx:
Addressed the three review points:

- **Items-per-page dropdown padding**: root cause was `--cds-layout-size-height-local` (which `.cds--list-box__menu-item__option`'s padding is computed from) only ever being set on `.cds--list-box`, while `ember-power-select` portals the dropdown content out of that subtree into `document.body` by default — so the property never inherited and every option rendered with 0 block padding. This is a pre-existing `Select`/`Pagination` bug, not specific to `AiChatTable` (repros on the plain `Pagination` docs page too). Fixed in `src/styles/index.scss` by re-declaring the same default formula on the wormhole's own `.cds--list-box__menu`.
- **CSV export**: `stringifyCSV`/`download()` were already producing correct, RFC 4180 comma-delimited output (verified byte-for-byte via a real download). "All columns in one column" is the classic Excel behavior of parsing CSVs using the OS/Excel locale's list separator instead of the file's actual delimiter (common outside en-US locales). Fixed by prepending a `sep=,` hint line — Excel's own documented escape hatch — in `AiChatTable.download()`, not in the shared `stringifyCSV` formatter (which stays a byte-identical, spec-compliant port of upstream's utility).
- **ChatShell input**: confirmed out of scope — `ChatShell` owns no input state, matching upstream (the real prompt input is a separate, not-yet-ported `prompt-line` widget). The docs demo was upgraded from a static placeholder to a real, typeable `TextInput` + `Button` so the example is honest about what a real integration looks like, but no component-level change was needed.

All verified with a real `DOCS_URL=versions/main` build served locally and driven with Playwright (screenshot + interaction), plus the full test-app suite (723/723 green).

patricklx:
Fixed: the "Confirm action" `AiChatCardFooter` demo's Accept/Reject buttons showed a stray light-gray/white triangle at their rounded outer-bottom corners in dark mode.

Root cause: `.cds-aichat-card-footer__actions`'s "grout" background between adjacent buttons is driven by `var(--cds-button-separator, #e0e0e0)`. Unlike this file's other `--cds-chat-*` tokens, `--cds-button-separator` is a button-component-specific custom property that isn't reliably set outside a real `.cds--g90`/`.cds--g100`-classed ancestor — so it silently falls back to its light `#e0e0e0` default even when the rest of the card (background, borders, button colors) is correctly dark-themed. Since only the buttons themselves got the rounded-corner treatment (not their container), the container's own square corner — filled with that wrong light color — peeked out from behind the button's rounded corner.

Fix: give `.cds-aichat-card-footer__actions`/`__icon-actions` the same corner radius as their rounded children (plus `overflow: hidden`), so the container is clipped to the same shape and its background can never show past the button's own corner, regardless of what color it resolves to. Verified visually via a local `DOCS_URL=versions/main` build in both light and dark (g90/g100) themes — corners are clean in both, no regression to the non-rounded corners or the icon-actions/danger-button variant.

PR #862, opened 2026-09-10
feat: port Carbon AI Chat ChainOfThought, ReasoningSteps, Feedback, PromptLine components

patricklx:
## Summary

Ports 4 of the 5 requested `@carbon/ai-chat-components` widgets (batch 2 of the Carbon AI Chat initiative) as Ember components under `src/components/ai-chat/`, following the same approach as PR #838 (Launcher/ChatShell) and the batch-1 PR #840 (Card/Table/TruncatedText):

- `ChainOfThought` + yielded `ChainOfThoughtStep`, `ChainOfThoughtToggle`
- `ReasoningSteps` + yielded `ReasoningStep`, `ReasoningStepsToggle`
- `Feedback`, `FeedbackButtons`
- `PromptLineShell`, `PromptLine` (**textarea mode only** — see below)

None of these collide with a Carbon React component name or an existing filename, so all eight export names stay unprefixed and no `AI_CHAT_EXPORT_OVERRIDES` entry was needed.

**`chat-history` was split into its own follow-up todo.** Its manifest has ~10 sub-elements (`history-shell`/`-header`/`-toolbar`/`-panel`/`-panel-items`/`-panel-item`/`-panel-item-input`/`-panel-menu`/`-search-item`/`-loading`/`-delete-panel`) — materially larger than any single "component" shipped so far — and wraps `CDSSideNav`/`CDSSideNavItems` plus a real overflow-menu repositioning concern, which deserves its own dependency review rather than being folded into this PR for size reasons.

**`PromptLine` ports the default textarea editing surface only.** Upstream's rich Tiptap editor (`@rich`, `@extensions`, `getEditor()`/`ensureEditor()`, `undo()`/`redo()`, `insertContent()`, `setTextSelection()`) is a genuinely new runtime dependency decision — the same reasoning `flatpickr` (DatePicker), `@carbon/utilities` (Carousel), and `markdown-it`/`dompurify` (Markdown) were each split out for their own review. None of the rich-mode args are accepted as silent no-ops; they're left out entirely and documented as a follow-up in the class doc.

`ChainOfThought`/`ReasoningSteps` steps are a genuine controlled/uncontrolled case (unlike Launcher/ChatShell, which had no open/closed concept at all upstream) — reproduced 1:1 with `@open ?? internalTracked` + a `!@controlled` write guard, and the container yields the step pre-bound with `@controlled` via `WithBoundArgs` (the `Layer` pattern).

Full write-up of all the scope decisions is in `AGENTS.md`'s "Porting Carbon AI Chat" section under the new "Batch 2" heading.

## Test plan

- [x] `pnpm build:js` / `pnpm build:types` (glint) — clean
- [x] `pnpm lint` (hbs/js/types) — clean (0 new errors/warnings; pre-existing warnings elsewhere untouched)
- [x] Full `test-app` suite: **804/804 green** (41 new integration tests across the 8 components)
- [x] Real `DOCS_URL=versions/main pnpm build`, served locally with an SPA-fallback static server, driven with Playwright: all 8 new `.gjs.md` demo pages render inside a real shadow root with populated content and zero page errors; interactive flows verified end-to-end (chain-of-thought toggle expands/collapses, reasoning-step disclosure, prompt-line typing + Enter-to-send + field clears, feedback category selection, feedback-buttons thumbs-down opens a Feedback panel)


patricklx:
Addressed review round 1 findings:

1. **Fixed:** `Feedback`'s `@initialValues` never seeded the panel on initial render. `did-update` explicitly doesn't fire on first render, so `applyInitialValues()` is now also called from the constructor; `didUpdate` is kept only for resetting on a later identity change. Added a regression test asserting the text area/selected categories are populated at mount.
2. **Fixed:** expanded AGENTS.md's "Batch 2" section with a `feedback`/`feedback-buttons` subsection covering the plain-button category-chip decision, `Markdown`-rendered disclaimer, `Tooltip`'s `@autoAlign` replacing upstream's direction-based flip, and the `did-update`-only-seeding gotcha above.

Full checklist re-run clean: `pnpm build:types` (glint), `pnpm build:js`, `pnpm lint` (0 errors), full `test-app` suite 805/805 green (+1 new test).

patricklx:
Review round 2 fixes pushed:

1. **`ChainOfThoughtToggle`/`ReasoningStepsToggle` permanently locked open state.** The `open` getter keyed on whether `@open` was defined rather than whether `@onToggle` was passed, so a static `@open={{true}}` seed (no `@onToggle`) got stuck after the first click. Fixed to follow this repo's own Case B convention (see `tree-view.gts`'s `TreeNode.expanded` / AGENTS.md): key on `@onToggle` presence, seed `internalOpen` from `@open` at construction. Updated both components' tests to cover the seed-then-click case, plus a proper fully-controlled test (`@open`+`@onToggle` wired together via `cell()`), replacing the old test that incorrectly assumed `@onToggle` alone drives self-toggling.

2. **`ChainOfThought`'s container `Args.onToggle` was declared/documented but never invoked.** Wired via a small modifier that calls `@onToggle` whenever `@open` changes after initial render (mirroring upstream's `updated()`-driven dispatch, excluding the initial mount). `ReasoningSteps` deliberately has no such arg upstream, so it's left as-is.

Full checklist re-run clean: glint, `build:js`, `lint` (0 errors), full test-app suite 810/810 green (5 new/updated tests).

patricklx:
Review round 4 fixes, pushed:

1. **`ChainOfThoughtStep` never collapsed its body content.** `hidden` was bound to `{{if (has-block) false true}}`, ignoring `this.isOpen` entirely, so any step with body content rendered permanently expanded. Fixed to `hidden={{if (and (has-block) this.isOpen) false true}}` (hidden unless there's a body *and* it's open), matching upstream's `!isHidden || !open`. Also dropped the dead `--open` modifier class the SCSS never styled. Added a regression test asserting the content starts `hidden`, opens on click, and re-collapses on a second click.

2. **`PromptLineShell`'s `@expanded` reflow destroyed/recreated the yielded editor.** Toggling `{{#if @expanded}}` around `{{yield to='editor'}}`/`{{yield to='messageActions'}}` tore down and rebuilt whichever branch was inactive, losing any live editor state (focus, selection) on toggle — contradicting the JSDoc's claim of an identical DOM. Fixed by rendering both blocks in one fixed DOM location and using CSS `order` (scoped under `--expanded`) to swap their visual sequence instead. Added a regression test asserting the same editor/message-actions DOM nodes survive an `@expanded` toggle, and updated the existing DOM-order test to reflect the new (correct) fixed ordering.

Full checklist green: `pnpm glint`, `pnpm build:js`, `pnpm lint` (0 errors), full `test-app` suite 813/813.

patricklx:
## Review round 6 fix — Feedback's didUpdate reintroduced @ember/render-modifiers

`Feedback` seeded/reset `@initialValues` via `{{didUpdate this.applyInitialValues @initialValues}}`, reintroducing `@ember/render-modifiers` into brand-new code against AGENTS.md's own "What NOT to Reach For" rule — and needing an extra constructor call as a workaround only because `did-update` never fires on initial render.

Fixed by replacing it with a real `ember-modifier` (`watchInitialValues`, same shape as `chain-of-thought.gts`'s `watchOpen`) that takes `@initialValues` as its tracked positional arg and calls `applyInitialValues()` — fires on both initial install and later identity changes, so the separate constructor call was deleted too. Also dropped the now-vestigial `@action` decorator on `applyInitialValues` (no longer called from a template mustache or callback reference).

Checked the rest of this PR's new files for the same pattern — none of them import `@ember/render-modifiers` (the two other hits in `src/components/ai-chat/` are `truncated-text.gts`/`table.gts`, both already on `main` from PR #838/#840, not part of this batch).

Verified: `pnpm exec glint`, addon `build:js`, `lint:js` all clean; full `test-app` suite 813/813 green both before and after the `@action` cleanup, including all 9 `ai-chat/Feedback` tests (the "seeds the text area and selected categories on initial render" test is what actually exercises this fix).

patricklx:
Fixed the three reported layout bugs, all verified visually with a real `DOCS_URL=versions/main` build served locally and driven with Playwright (before/after `getBoundingClientRect()` measurements, not just re-reading the CSS):

1. **PromptLineShell button "not aligned"**: the demo passed plain text `Send` into an `@iconOnly` Button with no icon — that rendered a cramped, oversized primary button with clipped text instead of the small ghost icon button upstream's real `send-control` renders (`kind="ghost" size="sm"` + a 16px send icon). Fixed the demo to match: `@ghost @size='sm'` + `<Send @size='16' />` + `aria-label`.
2. **ChatShell input button not aligned**: same underlying cause — swapped the ad hoc `TextInput` + `Button` combo for a real `PromptLineShell` + `PromptLine` (which now exists on this branch), with the same icon send button. This also fixes the doc's now-stale "hasn't been ported yet" claim.
3. **Toggle history pushes input out, toggle workspace doesn't**: `__history` is a fixed 320px column and `__input-and-messages` has a hardcoded 320px `min-inline-size` — together they need 640px and overflow any narrower shell. Upstream has the identical CSS and only avoids this via its (deliberately unported) `ResizeObserverManager` breakpoint collapse. Let `__input-and-messages` shrink below its floor once `@showHistory` is set, documented inline as a port decision — `@showWorkspace` alone never needed this since `__workspace` already had `min-inline-size: 0`.

Full `test-app` suite 813/813 green, glint/lint clean.

patricklx:
Addressed review round findings:

- **Must fix**: both new Send buttons (`chat-shell.gjs.md`, `prompt-line-shell.gjs.md`) passed `@ghost={{true}}` without `@type={{undefined}}`, so `Button`'s `@defaultArgs` kept injecting the `primary` type default alongside `ghost` - rendering `cds--btn--primary cds--btn--ghost` together (only looked correct by accident of CSS source order). Fixed both to explicitly pass `@type={{undefined}}`, matching the established `card-footer.gts` pattern for this exact situation.
- **Worth considering**: added a regression test (`chat-shell-test.gts`) for the `@showHistory` overflow CSS fix, rendering real `@carbon/styles` CSS and asserting the input column stays within the shell's bounds on a narrow container - guards against a future SCSS reorg silently reintroducing the overflow.

Verified: `glint`, addon `build:js`, `lint` all clean (0 errors); full `test-app` suite 814/814 green.

PR #863, opened 2026-09-11
feat: add rich (Tiptap) editing mode to PromptLine

patricklx:
## Summary

- Adds the Tiptap rich-editing mode that PR #862 (batch 2) deliberately left as a follow-up for `PromptLine`. New runtime dependencies: `@tiptap/core`, `@tiptap/pm`, `@tiptap/extensions` (for `UndoRedo`), and `@tiptap/extension-{document,paragraph,text,hard-break,placeholder}`, all pinned exact at `3.31.3` (same new-dependency review flatpickr/`@carbon/utilities`/markdown-it+dompurify each got).
- `PromptLine` still renders a plain `<textarea>` by default and never statically imports `@tiptap/*` — every real `@tiptap/*` import lives in one file (`ai-chat/-prompt-line/rich-controller.ts`) reached only through a dynamic `import()`, confirmed as its own separate chunk (`rich-controller-*.js`, ~176KB gzipped) in a real `DOCS_URL=versions/main pnpm build`, not bundled into the main entry.
- Setting `@rich` (or calling `ensureEditor()`) upgrades the textarea to the Tiptap surface in place, carrying text, caret, and focus over losslessly (verified in a real browser, not just unit tests).
- New `@onReady?: (api: PromptLineApi) => void` callback hands back a stable imperative handle (`getEditor`/`ensureEditor`/`undo`/`redo`/`insertContent`/`clearContent`/`setTextSelection`/`selectAll`/`focus`/`blur`/`hasFocus`/`getValue`) that works the same in either mode — no existing component in this addon had this pattern before, so it's now documented in AGENTS.md for reuse.
- `@extensions?: Extension[]` accepts host-supplied plain Tiptap extensions, appended to the base Carbon bundle (schema, undo/redo, placeholder, plain-text paste, the Enter/Mod-Enter/Escape keymap). Compared by reference (not upstream's deep equivalence check) — memoize the array.
- Deliberately still out of scope, documented in AGENTS.md: mention/autocomplete extensions (`carbon-mention`/`carbon-autocomplete`/`carbon-starter-trigger`), a typing-indicator event, and IME-composition guarding around the swap.

Full write-up of all the scope decisions (lazy-load verification, `@extensions` semantics, the lossless swap, the new `@onReady` pattern, and why the keyboard-vs-pointer focus-ring/typing/keydown gaps stay out of scope even with two editing surfaces now) is in AGENTS.md's "Porting Carbon AI Chat" → batch-2 `prompt-line` section.

## Test plan

[...]

PR #864, opened 2026-09-11
PromptLine: IME composition guard + Tiptap preload fast path

patricklx:
## Summary
Follow-up to PR #863, closing two of its three documented rich-mode scope cuts:

- **IME composition guard**: a composition in flight (typing e.g. Japanese/Chinese/Korean) now defers both a `@rich`-triggered upgrade and a rich-mode `@extensions` rebuild until composition ends, instead of tearing the field out from under an in-flight candidate. Ported from upstream's `_isComposing`/`_pendingUpgrade` guard and its rich-runtime withheld-recreate logic. Adds `setComposing()` to `EditingSurfaceController` (no-op in `TextareaController`).
- **Preload / warm-mount fast path**: a new `-prompt-line/rich-loader.ts` module shares one module-level Tiptap-chunk cache between the cold-start upgrade path and a new synchronous warm-mount check in `mountSurface`. `static PromptLine.preloadRich()` lets a host warm the chunk ahead of render, so an initial `@rich={{true}}` mounts the rich editor directly with no textarea-then-swap flash once warm.

The third cut (mention/autocomplete Tiptap extensions) is intentionally **not** ported here — reading upstream's real source surfaced enough reasons (a new `@tiptap/extension-mention` dependency, re-opening the transaction-origin-tagging cut this port deliberately made, a light-DOM portal rendering pattern with no precedent in this addon, and an open API-shape question) that it needs its own dependency-review todo, same as flatpickr/`@carbon/utilities`/markdown-it before it. Written up as a scoped decision in AGENTS.md; a follow-up todo will be filed for it.

Also revisits the documented keyboard-vs-pointer focus-ring gap in `PromptLineShell`'s expanded layout. AGENTS.md previously suggested a `:has(:focus-visible)` fix; verified directly in a real Chromium instance that this **cannot work**, since browsers already match `:focus-visible` on a plain mouse click for text-entry elements (`<textarea>`/`contenteditable`), unlike a `<button>`. AGENTS.md is corrected to describe the actual gap accurately (a second, keyboard-only wrapper outline upstream has that this port never added at all) and to explain why the CSS-only fix doesn't close it — reproducing it for real needs upstream's actual `MouseFocusController` event-plumbing, out of scope as a "small CSS fix."

## Test plan
- [x] `pnpm exec glint` — clean
- [x] `pnpm run build:js` (addon) — clean
- [x] `pnpm run lint` (addon: js/hbs/types) — 0 errors, only pre-existing unrelated warnings
- [x] Full `test-app` suite — 838/838 green (835 + 3 new: two IME-guard tests, one preload/warm-mount test)
- [x] Real `DOCS_URL=versions/main pnpm build`, served locally with an SPA-fallback static server, driven with Playwright:
  - Confirmed `rich-controller-*.js` is still its own separate lazy chunk (~178KB gzip) — the split is intact
  - Dispatched a real `compositionstart` on the textarea, clicked "Switch to rich editing": upgrade stayed deferred (textarea still present, no `pm-content`) until a `compositionend` was dispatched, at which point the swap completed
  - Confirmed the textarea→rich swap still carries text/focus over losslessly after these changes
  - A pre-existing, repo-wide `TypeError: Cannot read properties of undefined (reading 'getChildByName')` page error was observed — confirmed unrelated to this change by reproducing it on the completely unrelated `button` docs page too

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1276,11 +1276,7 @@ upstream's own `TextareaController`/`RichController` pair
 (`prompt-line-controller.ts`/`prompt-line-rich-runtime.ts`) — upstream's
 classes are already plain DOM/Tiptap code with no Lit dependency, so this
 port is closer to a straight port than a re-derivation. Real cuts made
-against upstream, all deliberate: no mention/autocomplete extensions
-(`carbon-mention`/`carbon-autocomplete`/`carbon-starter-trigger`/
-`token-node-view` — a separate, not-yet-ported feature; `@extensions`
-accepts plain Tiptap `Extension`s only, so no `content`-as-`JSONContent`
-seeding either, only plain strings), no typing-indicator event (nothing in
+against upstream, all deliberate: no typing-indicator event (nothing in
 this port consumes one — a caller can debounce `@onChange` itself), no
 origin-tagging on transactions (exists upstream only to keep the
 mention-removal plugin and the typing indicator from misfiring on a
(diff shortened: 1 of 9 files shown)


patricklx:
Review round 1 fix: `mountSurface`'s teardown reset `isComposing`/`pendingUpgrade` but never called `failRichReady()` when a composition-deferred upgrade was still pending - if the component was destroyed while an `ensureEditor()` caller was awaiting a rich-mode upgrade that never got to fire (composition started, never got a `compositionend`), that caller's promise hung forever. Fixed to fail `richReadyPromise` in teardown when it's still outstanding, matching the sibling mid-`loadRichRuntime()` teardown path that already handled this. Added a regression test (destroy while a composition-deferred upgrade is pending, assert `ensureEditor()` rejects rather than hangs).

Full checklist re-run clean: glint, addon `build:js`, addon `lint` (0 errors), full `test-app` suite 839/839 green (up from 838).

patricklx:
Fixed the review round 2 finding: `rich-loader.ts`'s module-level Tiptap runtime cache was never reset between tests, so once one test warmed it, every later `@rich={{true}}`-from-start test in `prompt-line-test.gts` silently took the new warm-mount fast path instead of the cold textarea-then-upgrade path they were originally written to exercise.

Added a `resetRichRuntimeForTests()` test-only export from `rich-loader.ts` and call it from an `afterEach` hook, restoring cold-path coverage for every pre-existing test. Full checklist re-verified: glint clean, addon `build:js`/`build:types` clean, `lint` 0 errors, full `test-app` suite 839/839 green, and a real `DOCS_URL=versions/main pnpm build` served locally + driven with Playwright confirms both `prompt-line`/`prompt-line-shell` docs demos still render inside a real shadow root with zero page errors and working input.

PR #865, opened 2026-09-11
Carbon AI Chat batch 3: Processing, ChatButton, Toolbar, WorkspaceShell, FileUploads

patricklx:
## Summary
- Ports 5 more `@carbon/ai-chat-components` widgets as Ember components: `Processing`, `AiChatChatButton` (+skeleton), `Toolbar`, the `WorkspaceShell` family (+Header/Body/Footer), and the `FileUploads` family (+FileUploadItem).
- `Toolbar` and `WorkspaceShellFooter` reuse this addon's own `Button`/`Tooltip`/`OverflowMenu` instead of `@carbon/web-components`; `FileUploads`/`FileUploadItem` render their own markup (reusing the private `FileUploaderStatusIcon`) instead of wrapping `cds-file-uploader-item`.
- `code-snippet` (needs a CodeMirror 6 dependency review - 24 `@codemirror/lang-*` packages) and `audio-player`+`video-player` (share third-party SDK-loading provider infrastructure) were split out into their own follow-up todos rather than pulled in silently, matching the precedent set by carousel/markdown/flatpickr in earlier batches.
- `chat-button` collides with Carbon React's own not-yet-implemented `ChatButton` - exported as `AiChatChatButton`, with the override added to both `scripts/parity-check.mjs` and `create-files.mjs`.
- Full design rationale (including two real bugs found and fixed along the way - a `ResizeObserver` feedback loop in `Toolbar`'s overflow measurement, and a backtracking-rerender bug in `FileUploads`' live-region announcer) is written up in `AGENTS.md`'s "Porting Carbon AI Chat" section under "Batch 3".

## Test plan
- [x] `pnpm exec glint` - clean
- [x] addon `pnpm run build:js` - clean
- [x] addon `pnpm run lint` - 0 errors (only pre-existing, unrelated warnings)
- [x] full `test-app` suite - 882/882 green
- [x] real `DOCS_URL=versions/main pnpm build`, served locally, driven with Playwright across all 10 new docs pages - populated shadow roots, zero unexpected page errors, and targeted interaction checks (chat-button quick-action selection, toolbar overflow menu, file-uploads remove, workspace-shell-header collapse toggle) all verified working in a real browser


patricklx:
Fixed all review round 1 findings:

- **`chat-button.gts`**: `buttonType` now only maps `'ghost'`/`'tertiary'` to `undefined`, so `@kind='danger'` renders with the correct `cds--btn--danger` class instead of no type class at all.
- **`-toolbar-action.gts`**: a disabled `href` action now strips `href`, adds `role="link"`/`aria-disabled="true"`, and blocks the click via a handler - matching `link.gts`'s established pattern. It's no longer navigable while disabled.
- **`workspace-shell.gts`**: `observeCollapse` now defers its `shouldCollapseHeader` write to `requestAnimationFrame`, matching `Toolbar`/`WorkspaceShellFooter`'s existing ResizeObserver-loop-avoidance pattern.
- **`toolbar.gts`**: `recomputeOverflow` now clamps its cutoff index to never go below the count of `fixed: true` actions, so fixed actions can no longer get pushed into the overflow menu when space is tight.

Added regression tests for all four. Full checklist green: glint, addon `build:js`, `lint` (0 errors), full `test-app` suite 887/887.

patricklx:
Review round 3 fix: `Toolbar`'s `displayedVisibleActions` getter always returned `sortByFixed()`'s output, which moves `fixed: true` actions to the front regardless of `@overflow` — contradicting the `@overflow={{false}}` (default) JSDoc contract that actions render "in their given order, with no measurement and no overflow menu." Confirmed against upstream: fixed-first ordering upstream only ever applies once overflow measurement actually runs.

Fixed to return the raw `@actions` array when `@overflow` is unset, keeping `sortedActions` for the `@overflow`-enabled measurement/render path only. Added a regression test asserting DOM order matches array order with a `fixed` action not listed first and `@overflow` unset.

Full checklist green: glint, addon `build:js`, `lint` (0 errors, only pre-existing unrelated warnings), full `test-app` suite 890/890.

patricklx:
Review round 4 fixes pushed:

- **`WorkspaceShellFooter.sortedActions` tied-rank stacking bug (real):** the stacked branch sorted ascending then `.reverse()`d the whole array, which also flips the relative order of same-kind (tied-rank) actions - `Array.prototype.sort`'s stability only preserves tie order under a single sign-flipped comparator (what upstream's real `_sortActions` does), not under sort-then-reverse. Fixed to mirror upstream exactly: one sort with the comparator's sign flipped by `isStacked`.
- Added regression tests for the stacked/narrow reorder behavior (previously untested at all - none of the existing tests rendered the footer under 671px) and specifically for the tied-rank case the bug above got wrong.
- Added a test for `FileUploadItem`'s `SUCCESS` status (checkmark) branch, which had no coverage.

Full checklist re-run clean: `pnpm exec glint` (0 errors), addon `build:js`, addon `lint` (0 errors, pre-existing warnings only), full `test-app` suite 893/893 green.

patricklx:
Review round 5 fixes pushed:

- `file-upload-item.gts`: default `uploadingFileLabel` fallback changed from `'Uploading'` to `'Uploading file'`, matching upstream's real default exactly.
- `workspace-shell-footer.gts`: dropped the dead, unreachable `'danger--ghost'` entry from the `RANK` sort table (the `kind` type doesn't support it and `Button` can't render that combination class), and documented the gap in the class doc the same way `AiChatChatButton` already does for the identical combination-kind limitation.

Full checklist re-verified green: glint, addon `build:js`, `lint` (0 errors), full `test-app` suite 893/893.

patricklx:
Fixed the round-6 finding: `FileUploadItem.getOrCreateObjectURL` read the `@tracked objectURL` field (in its revoke check) and then wrote it in the same render-time computation, which tripped Ember's backtracking-rerender assertion the instant any image/video-typed upload rendered — not just on update, on first paint. Same category of bug as `FileUploads.activeRegion` earlier in this PR: `objectURL`/`objectURLFile` are now plain (non-`@tracked`) fields, since nothing outside `getOrCreateObjectURL` reads them directly — only `previewURL`'s return value, whose reactivity already traces to `@upload`.

Added image/video preview test coverage (previously untested), a same-instance `@upload`-identity-change test asserting the stale object URL is revoked exactly once, and gave both file-upload docs demos a real (SVG) image so the path is exercised in a live browser too — verified via a local `DOCS_URL=versions/main` build + Playwright that the preview actually decodes (`naturalWidth: 36`), not just carries the right MIME type. Full checklist re-run clean: glint, addon build, lint (0 errors), full test-app suite 896/896.

patricklx:
Fixed the strange-looking close button on file upload chips: `AiChatFileUploadItem` reuses `FileUploaderStatusIcon` (the shared close/loading/complete button), whose `.cds--file-close`/`-loading`/`-complete` classes are only styled by `@carbon/styles` when nested under `.cds--file__state-container`. `file-uploader-item.gts`'s own usage already wraps with that class; this port's `__status` span never did, so the close button rendered with default native `<button>` chrome (border, background, padding) instead of a clean 24x24 transparent icon button.

Fix: add `cds--file__state-container` alongside the existing `cds-aichat-file-upload-item__status` class. Verified via a real `DOCS_URL=versions/main pnpm build` served locally + Playwright — the button now computes to `border: 0px none`, transparent background, 24x24, `cursor: pointer`. Added a regression test asserting the class is present. Full checklist green: glint, addon `build:js`, addon `lint` (0 errors), full `test-app` suite 949/951 (the 2 failures are the pre-existing `Grid` style-snapshot env flake, unrelated).

patricklx:
Fixed the red `test` check (pushed to `feat/ai-chat-batch3`).

Root cause: `evaluateShowMoreButton`'s `shouldCollapse` check (ported verbatim from upstream) never guarded against its own threshold (`minExpandedNumberOfRows`, default 16 rows) overlapping `shouldShowButton`'s threshold (`maxCollapsedNumberOfRows`). The existing test passes `@maxCollapsedNumberOfRows={{3}}`, opening a wide band where content both needs the show-more affordance and is "small enough" to auto-collapse — clicking "Show more" immediately snapped back to collapsed. This is a genuine latent bug present in upstream too (confirmed identical logic there), just not exercised by upstream's own tests. It only reproduced under CI's font-metric environment, not locally, which is why it read as a flake at first.

Fixed by making the two conditions mutually exclusive (`shouldCollapse` now also requires `!shouldShowButton`). Added a deterministic unit test (mocked container height, no real-DOM font dependency) plus verified against 3 real CI runs before/after.

Confirmed via a fresh CI run on the pushed commit: `AiChatCodeSnippet` and `DataTable` are now fully green — only the two pre-existing, already-documented `Grid` snapshot flakes remain (env font-metric drift, unrelated to this branch, also reproduces on `main`).

patricklx:
Fixed: **AiChatChatButtonSkeleton docs page was not showing anything.**

Root cause: `cds--skeleton` has no standalone CSS rule of its own in `@carbon/styles` — it only exists as a compound modifier paired with a specific component's own class (`.cds--btn.cds--skeleton`, `.cds--skeleton.cds--text-input`, `.cds--skeleton.cds--tabs`, etc.). The component applied it alone to a bare `<div>`, which rendered with zero background/animation — an invisible placeholder, not a broken import/build.

Fixed by pulling in `@carbon/styles/scss/utilities/skeleton`'s `skeleton` mixin directly (same approach Carbon's own component skeletons use under the hood) scoped to `.cds-aichat-button-skeleton`, rather than the class or the generic `cds--skeleton__placeholder` utility (which ships its own default block-size/inline-size at equal specificity to our own sizing rules, which would have created a cascade-order tie).

Verified for real, not just by reading CSS: compiled the modified SCSS partial directly and confirmed the `::before` shimmer + background rules land correctly with no specificity conflicts; then did a real `DOCS_URL=versions/main pnpm build` served locally + Playwright pass confirming all three size variants (`lg`/`md`/`sm`) render with correct dimensions, a visible background color, and the animated shimmer `::before`, zero page errors.

Full checklist: `pnpm exec glint` clean, addon `build:js` clean, addon `lint` (0 errors), full `test-app` suite 954/954 green on a clean rerun (an earlier run showed 3 failures — 2 confirmed pre-existing Grid snapshot flakes plus 1 unprinted due to this repo's known "test-list caps around #850" harness quirk; a full rerun came back 954/954, confirming none were caused by this change).

patricklx:
CI is green now (`test`, all 3 `ember-try` scenarios, `Deploy Preview`, `DCO`).

Root cause: the Grid style-snapshot fixture (`white theme`/`dark theme: should display grid`) was stale, not a real regression. Bisected via scratch branches: `origin/main` passes on today's runner, and the commit right before this branch's CodeMirror/AiChatCodeSnippet merge also passes — every run since that merge failed deterministically. No CSS selector collision from CodeMirror's (all `.cm-`-scoped) injected styles was found; the stored snapshot encoded a layout where both `<GridColumn>`s never actually got Carbon's `.cds--row` flex rules applied (full-width, stacked, 32px tall) — a latent CSS-application-timing race in the Grid test itself that predates this PR and was only newly exposed by CodeMirror's addition shifting stylesheet-injection timing earlier in the same shared test document. CI's current rendering (side-by-side, 622px each, matching Carbon's actual default column-split behavior) is the correct one. Regenerated the fixture via a scratch CI run (`pnpm run test:ember:update-snapshot`, since this doesn't reproduce locally on macOS) and applied the values directly.

Also: the previously-reported "3 failures" in earlier CI runs was 2 real Grid failures + 1 phantom count from a `js-reporters` quirk (it double-counts a module as its own "test" when any test inside it fails) — no hidden 3rd bug.

PR #868, opened 2026-09-12
PromptLine: add the real suggestion popup (PromptLineAutocomplete)

patricklx:
## Summary
- Ports `@carbon/ai-chat-components`' `autocomplete-controller.ts` + `prompt-line/autocomplete/src/autocomplete.ts` as one new component, `PromptLineAutocomplete` — the suggestion popup for `PromptLine`'s mention/command/autocomplete/starter Tiptap extensions that PR #866 deliberately left unported.
- Fixes the documented same-transaction race between independent trigger extensions for real: batches `cds-aichat-trigger-change` events by microtask and resolves each batch to its last non-null detail (upstream's own `AutocompleteController.handleTriggerChange` is last-event-wins and reproduces this race rather than fixing it — verified by reading the fetched source, not assumed).
- Selection/dismissal reuse `PromptLineApi.selectSuggestion()`/`dismissSuggestion()` (added in the previous PR precisely for this) instead of re-deriving the insertion logic; event wiring resolves its listen target from DOM ancestry rather than `@promptLine` as a modifier dependency, avoiding a backtracking-rerender hazard.
- Recombines the docs demo split across two `PromptLine` instances back into one editor with mention+command+starters together, as the live proof of the fix.

## Test plan
- [x] `pnpm exec glint` / addon `build:js` / addon `lint` (hbs/js/types) all clean
- [x] Full `test-app` suite green (872/872), including a reconciliation matrix (all four constructible same-tick batch sequences) and a real mention+starters integration test
- [x] Real `DOCS_URL=versions/main pnpm build` served locally + Playwright: both new docs pages render inside a real shadow root with zero page errors; confirmed live end-to-end (focus → starters show → typing `@` → mention items show, not closed; click → chip inserted; groups + icon avatars render correctly)


patricklx:
Rebased and fixed CI.

**Rebase:** PR #866 (this branch's original base, `feat/prompt-line-mentions`) merged into `main` as #866 mid-task, and GitHub auto-retargeted this PR's base to `main` once that branch was deleted. Rebased with `git rebase --onto origin/main <old-fork-point> HEAD` to replay only this PR's own 4 commits onto the new `main` tip (dropping the now-redundant commits already squashed into #866's merge) — a plain rebase briefly produced `CONFLICTING`/duplicate-content merge conflicts by trying to replay commits whose content already landed via the squash-merge.

**CI fix — real bug in already-shipped code, not this PR's own diff:** `rich-controller.ts`'s `createChatEnter`'s `Enter` handler (and `createChatKeymap`'s `Mod-Enter` handler) declined to handle (`return false`) a keydown while a mention/command/autocomplete trigger was active, intending to leave the keystroke for a host popup. But nothing else in the extension bundle binds plain `Enter`, so an untrapped `Enter` fell through to the browser's default contenteditable paragraph-split behavior, and an untrapped `Mod-Enter` fell through to `HardBreakNode`'s own unconditional `Mod-Enter -> setHardBreak()` binding — both insert content that desyncs the query from the cursor and exits the trigger. A second keydown right after (e.g. `Mod-Enter` immediately following an already-corrupted `Enter`) then saw `hasActiveSuggestion() === false` and genuinely sent, reproducing deterministically in CI as `"plain Enter and Mod-Enter do not send while a mention trigger is active"` failing (same failure on #866's own CI). Fixed by swallowing (`return true`) instead of declining while a trigger is active — this still lets a host popup see the raw keydown via DOM bubbling (no `stopPropagation`), just without letting ProseMirror's own fallback handlers act on it.

Split the one combined regression test into two isolated ones (plain Enter alone, Mod-Enter alone), each also asserting the document/trigger state survives, not just that nothing was sent. Note: the isolated Mod-Enter case doesn't reproduce against a local `pnpm test` run (environment/timing-sensitive), only on CI — verified the fix against 3 successive real CI runs before landing it, not just a local pass, since an earlier attempt at narrowing the fix based on a misleading local-only result had to be reverted.

PR #671, opened 2026-07-25
feat: implement TimePickerSelect component

patricklx:
## Summary

- Implements `TimePickerSelect`, a thin native `<select>` wrapper matching Carbon React's `TimePickerSelect` (`cds--select`/`cds--time-picker__select` wrapper, `cds--select-input` select, `ChevronDown` arrow icon).
- Supports `@id` (auto-generated if omitted), `@disabled`, `@defaultValue` (uncontrolled), `@value`/`@onChange` (controlled), and `@ariaLabel` (defaults to `"open list of options"`, matching React's default).
- `<option>` elements are passed as block content, matching React's `children` prop.

## Notes

- `TimePickerSelect` has no dedicated React stories of its own — it's only demonstrated as a subcomponent within `TimePicker`'s stories (AM/PM and timezone selects). Docs cover default usage, `defaultValue`, controlled value, and disabled state as standalone examples, plus an illustrative (non-live) snippet showing intended usage alongside `TimePicker` once that component exists.
- `TimePicker` itself is not yet implemented in Ember; this PR only covers `TimePickerSelect`, which doesn't depend on it.
- Splattributes are applied to the inner `<select>` (not the wrapper `div`), matching React's behavior of spreading rest props onto the native select and required to satisfy the `require-input-label` template-lint rule (an element with both `id` and `aria-label` and no splattributes trips "multiple labels").

Closes #502

## Test plan

- [x] `pnpm build` succeeds (types + js)
- [x] `pnpm lint` passes (0 errors)
- [x] New integration tests in `test-app/tests/components/time-picker-select-test.gts` cover: default rendering, id generation, aria-label default/override, disabled, defaultValue, controlled value, and onChange
- [x] Full test suite passes (272 passed, 0 failed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #877, opened 2026-09-14
Carbon AI Chat orchestration layer: carbon.ai-chat-session service + SessionShell

patricklx:
## Summary

Ember-native equivalent of `@carbon/ai-chat`'s React app (`packages/ai-chat/src/chat`: `AppShell` + `services/` + the Redux `store/`) — the orchestration/state-machine layer this repo's AI Chat port has deferred since todo #662 (rescheduled #671 → #721 with a mechanical readiness gate: PR #865 merged and `ai-chat/chat-history` genuinely ported — both confirmed before starting).

Two new files:
- `src/services/ai-chat-session.ts` — `ChatSessionService`, registered under this addon's existing `carbon.*` service namespace (same as `carbon.notifications`/`carbon.dialog-manager`) via `@service('carbon.ai-chat-session')`.
- `src/components/ai-chat/session-shell.gts` — `SessionShell`, the `AppShell`-equivalent container. **The one and only injection point**: it injects the service and passes session state down as plain `@arg`s to the already-ported `Launcher`/`ChatShell`/`PromptLineShell`/`PromptLine`/`Processing` components, none of which change — they stay exactly as shipped (stateless/always-controlled per their own class docs). This satisfies "not a parallel prop-drilling scheme": one injection point, one level of args, per AGENTS.md's "React context → a service, or the parent component instance yielded down to children" convention.

This is a deliberately scoped subset, not a 1:1 port — upstream's own orchestration layer is enormous (`store/reducers.ts` alone is 1,691 lines with ~50 Redux action types; `services/ChatActionsImpl.ts` is 2,399 lines; `AppShell.tsx` is 1,039 lines). Scope and cuts below, and in AGENTS.md's "Porting Carbon AI Chat" → "Orchestration layer" section.

## React → Ember mapping

| Upstream (React/Redux) | This port |
| --- | --- |
| `ServiceManager` (upstream's own DI container) | **Deleted entirely** — Ember's owner/`@service` already is this. |
| `store/` (Redux: `actions.ts`/`reducers.ts`/`appStore.ts`/`selectors.ts`) | `ChatSessionService`'s own `@tracked` fields (`messages`, `draft`, `open`, `showHistory`, `showWorkspace`, `isReadonly`) — no separate action/reducer/selector layers; methods mutate directly and Glimmer's autotracking replaces `connect()`/selectors. |
| `MessageService`/`OutboundMessageCoordinator` (`ADD_MESSAGE`, `UPSERT_MESSAGE`, `UPDATE_MESSAGE`) | `ChatSessionService#send()`/`#receive()` |
| `InboundStreamingCoordinator` (`STREAMING_START`, `STREAMING_ADD_CHUNK`, generation tracking, abort controllers) | `ChatSessionService#appendChunk()`/`#finalizeStreaming()` — generation tracking/cancellation cut, see below. |
| `instance.on()`/`.off()`/`.once()` (upstream's `serviceManager.eventBus`) | `ChatSessionService#on()`/`#off()`/`#emit()` — a plain `Map<type, Set<handler>>`; `.once()` not ported (no caller needed it yet). |
| `instance.send()` | `ChatSessionService#send()` — same host boundary: appends the user message and emits `'send'`; producing the assistant reply is the host application's job (upstream's `customSendMessage`), not this service's. |
| `SET_HISTORY_PANEL_OPEN`/`SET_WORKSPACE_PANEL_OPEN` | `ChatSessionService#toggleHistory()`/`#toggleWorkspace()` |
| `SET_VIEW_STATE`/`instance.changeView()` | `ChatSessionService#toggleOpen()` (a single `open` boolean — upstream's richer multi-view `ViewState` isn't needed since `Launcher`/`ChatShell` are the only two views this port has). |
| `RESTART_CONVERSATION` | `ChatSessionService#restart()` |
| `AppShell.tsx` (assembles `Launcher`/`MainWindow`/panels, owns the Context providers) | `SessionShell` — the single injection point described above. |

## Deliberately NOT ported (real, separable follow-ups — not oversights)

- **Human-agent handoff** (`humanAgentActions.ts`/`humanAgentReducers.ts`, `HistoryService`'s agent-transfer paths) — a whole second conversation mode with no presentational components ported yet to drive it.
- **Persistence/rehydration** (`UserSessionStorageService`, `persistenceUtils.ts`, `HYDRATE_CHAT`/`HYDRATE_MESSAGE_HISTORY`) — no storage backend decision made yet; messages are in-memory only and vanish on reload.
- **`CustomPanelManager`/`CustomPanelInstance`** — `SessionShell`'s `<:history>`/`<:workspace>` blocks cover the two panel *slots* `ChatShell` exposes, but not upstream's generic multi-panel-instance management. Yielded outward so a future `ai-chat/chat-history` (PR #870) integration can plug into `<:history>` without this component changing.
- **`NamespaceService`** (multi-instance isolation) — the service is a per-app Ember singleton; multiple independent chat widgets on one page need a real design not needed for a first pass.
- **`ThemeWatcherService`** — docs-app's own `ThemeSupport` already owns theming here; no gap to fill.
- **The bulk of `actions.ts`'s ~50 action types** — file uploads, structured data, nested messages, iframe/conversational-search/response panels, catastrophic-error state, disclaimer acceptance, message-option-selection UI state: real upstream concerns with no Ember consumer yet. Add the service method + field alongside a real consumer, not speculatively.
- **Cancellation/generation tracking** inside streaming — `appendChunk()`/`finalizeStreaming()` assume a single, uncancellable stream per message id.

## Other notes

- `on()`/`off()` are manual, matching upstream — `carbon.ai-chat-session` is an app-wide singleton, so a consumer registering a listener owns unregistering it (e.g. `registerDestructor`). The docs demo does this; a real leaked-listener-across-remount bug was caught and fixed before merge, with a regression test (`session-shell-test.gts`) covering exactly this contract.
- Message shape is intentionally minimal (`{ id, role, text, streaming? }`), a flat reassigned array (not `TrackedArray`/a normalized map), matching this initiative's established `@tracked` reassignment pattern. There's no upstream "message bubble" component to port (that's app-internal, not part of `@carbon/ai-chat-components`), so `SessionShell` renders plain, minimally-styled bubbles.

## Test plan

- [x] `pnpm exec glint` clean
- [x] addon `build:js` clean
- [x] addon `pnpm run lint` — 0 errors (1 real `{{#unless}}...{{else}}` lint error caught and fixed)
- [x] `pnpm run create-autogenerated-files` diff-checked — `SessionShell` needs no export-override entry; script's unrelated reordering reverted per established regen-gotcha
- [x] full `test-app` suite: 970/973 (3 = the pre-existing, already-documented Grid snapshot env-flake + its js-reporters module double-count — confirmed unrelated)
- [x] new service unit tests (send/receive/streaming/panel-toggle/event-bus/restart) and `SessionShell` integration tests (launcher↔shell toggle, real typed send via Enter, send button, streaming indicator, arg forwarding, readonly disabling, mount/destroy/remount listener-leak regression)
- [x] real `DOCS_URL=versions/main pnpm build` + Playwright pass on the new docs demo — launcher→shell toggle, real typed send, simulated streamed reply, close-back-to-launcher, zero page errors


patricklx:
Addressed review round 1 findings:
- Dropped the redundant `{{if (eq message.role "user") ...}}` class clause in `session-shell.gts` (the `--{{message.role}}` interpolation already produces the same token), and removed the now-unused `eq` import.
- Added a forwarding test asserting `@showHistory`/`@showWorkspace` on `ChatSessionService` reach `ChatShell` as `show-history`/`show-workspace` classes.

Verified: `glint` clean, addon `build:js` clean, `lint` 0 errors, full `test-app` suite 974/974 green.

patricklx:
Review round 2 fixes pushed (`37de21bb`):

1. **Streaming DOM node destroyed/recreated per chunk** — `session-shell.gts`'s `{{#each session.messages}}` had no `key=`, so `appendChunk()`'s new-object-reference-per-chunk defaulted to `@identity` keying and tore down/rebuilt the streaming bubble's DOM node on every chunk. Fixed with `key='id'` (message ids are stable across `appendChunk`/`finalizeStreaming`). Verified empirically: added a regression test asserting DOM node identity survives two `appendChunk()` calls, confirmed it fails without the fix and passes with it.
2. **`isReadonly` test didn't verify "prevents sending"** — only asserted the send button was disabled. Added the missing assertion (Enter while readonly leaves `session.messages` empty).
3. **Minor:** `emit()` now snapshots the listener Set before iterating, so a handler calling `on()`/`off()` mid-dispatch can't mutate it during iteration.
4. **Minor:** reworded the "one, and only, injection point" class doc — the `<:history>`/`<:workspace>` slots are a deliberate exception where a filler needs its own service injection.

Full checklist clean: `pnpm exec glint`, addon `build:js`, addon `pnpm run lint` (0 errors), full `test-app` suite 972/975 green (only the 2 pre-existing documented Grid style-snapshot flakes + their js-reporters module-aggregate double-count, unrelated).

PR #879, opened 2026-09-16
Carbon AI Chat: ChatSessionService persistence + chat-history integration

patricklx:
## Summary

Follow-up to #877 (`carbon.ai-chat-session` service + `SessionShell`) and the cancellation follow-up. Picks up two of the scoped cuts from #877's own "Deliberately NOT ported" list:

- **Persistence/rehydration**: `ChatSessionService` gains `enablePersistence(storage?, key?)` / `persist()` / `clearPersistedSession()` / `disablePersistence()`. Storage backend is a host-provided `ChatSessionStorage` interface (`Pick<Storage, 'getItem' | 'setItem' | 'removeItem'>`), defaulting to `window.sessionStorage` - matching upstream's own `UserSessionStorageService`, which deliberately avoids `localStorage` to sidestep an EU cookie-policy notice. A version stamp discards an incompatible persisted shape; a message persisted mid-stream restores with `streaming` cleared; the message-id counter is bumped past any restored ids to avoid collisions.
- **`ai-chat/chat-history` integration (#870)**: `SessionShell`'s `<:history>` block is no longer a pure passthrough - it assembles a real `ChatHistory` panel (header/toolbar/content/panel/items) from a new host-owned `@historyItems` arg plus session-owned open/close/new-chat chrome, matching upstream's own `customLoadHistory` host-boundary rather than inventing a `conversations[]` store inside the service. A `<:history>` block still overrides the whole thing when passed.

Full design rationale (why sessionStorage not localStorage, the session-owned/host-owned split, sanitize-on-restore steps, debounce choices) is written up in `AGENTS.md`'s "Porting Carbon AI Chat" → "Persistence and the `chat-history` integration" section.

## Test plan

- [x] `pnpm exec glint` (addon) - clean
- [x] addon `pnpm run build` (types + js) - clean
- [x] addon `pnpm run lint` (hbs/js/types) - 0 errors
- [x] full `test-app` suite - 1048/1051 (76 new tests; the 3 "not ok" are the pre-existing, already-documented `Grid` snapshot env-flake + its `js-reporters` module-aggregate double-count, confirmed unrelated by a clean rerun)
- [x] real `DOCS_URL=versions/main pnpm build` served locally + Playwright against the live `session-shell.gjs.md` demo: sent a message → reloaded the page for real → confirmed the shell stayed open and the message survived → opened the history panel, renamed and deleted a demo item via the overflow menu and delete-confirm overlay → "New chat" cleared the conversation and closed the panel - zero real page errors, and a `getBoundingClientRect()` check confirming the delete-confirm overlay is scoped to the history column, not the whole shell


patricklx:
Fixed the review finding: `renamingId`/`deletingId` now reset whenever the history panel closes, via a teardown modifier attached to the default `ChatHistory` assembly (so it fires regardless of which path closed the panel - header toggle, shell close, or `newChat()` - rather than intercepting each call site).

Added 2 regression tests, both verified to fail on the pre-fix code and pass after: one for the delete-confirm overlay (via the header toggle button, matching the review's exact repro), one for the rename input (via `svc.toggleOpen()` directly, since closing via a real button click also happens to blur the auto-focused rename input and trigger its own pre-existing, unrelated cancel-on-blur path - isolating the actual `renamingId` leak needed bypassing that).

Verified: `pnpm exec glint`, addon `build:js`, addon `pnpm run lint` all clean (0 errors); full `test-app` suite 1050/1052 (only the pre-existing, already-documented `Grid` white/dark-theme style-snapshot env flake, confirmed unrelated).

patricklx:
Review round 2 fix: `enablePersistence()` unconditionally discarded any pending debounced write (from `setDraft()`/`appendChunk()`) and rehydrated from the last-*flushed* snapshot even when called again with the exact same `storage`/`key` — the documented SPA-remount pattern. Fixed to flush the pending write instead of discarding it when the target is unchanged; a real change of `storage`/`key` still cancels the old pending write and rehydrates from the new target as before.

Added a regression test covering the same-key re-enable case, and documented the fix in AGENTS.md's "Persistence and the `chat-history` integration" section. Verified: `pnpm exec glint`, addon `build:js`, addon `lint` (0 errors), full `test-app` suite 1054/1054 green.

PR #881, opened 2026-09-16
fix: chat-history overflow-menu icon alignment + New chat button color

patricklx:
## Summary
- The Rename/Delete icons inside a `ChatHistoryPanelItem`'s overflow menu rendered below their label text instead of next to it. The item text and icon share one inline flow inside `.cds--overflow-menu-options__option-content`, which relies on the icon's `<svg>` keeping its browser-default inline `display`. Any host app with a common CSS reset that blockifies media elements (e.g. Tailwind's Preflight, confirmed present in `docs-app` itself) splits the icon onto its own line. Forced the icon back to `inline-block` in `src/styles/index.scss` (same "wormhole-target gap" pattern already used there for the `Select`/`Pagination` dropdown-padding fix), regardless of what a consumer's own reset does.
- `ChatHistoryToolbar`'s "New chat" button was rendered with `@ghost={{true}}`, overriding `Button`'s own `primary` default and rendering without a blue background. Upstream's `cds-icon-button` has no `kind` override and defaults to primary - dropped the explicit ghost override to match.

Verified with a real `DOCS_URL=versions/main pnpm build` served locally + Playwright: the Rename/Delete icons now sit on the same line as their label, and "New chat" renders with the primary blue background.

## Test plan
- [x] `pnpm exec glint` (addon) - clean
- [x] `pnpm run lint` (addon) - 0 errors, only pre-existing unrelated warnings
- [x] Full `test-app` suite - 1031/1031 green
- [x] Real `DOCS_URL=versions/main pnpm build` served locally + Playwright verification of both fixes in the `ai-chat/chat-history` docs demo


patricklx:
Addressed the three reported issues, all verified with a real `DOCS_URL=versions/main pnpm build` served locally + Playwright:

1. **Icons in the overflow-menu dropdown now align right.** `OverflowMenuItem` renders an icon as its own `.option-icon` sibling of the text, with `justify-content: space-between` on the trigger button - matches upstream Carbon web-components' flush-right icon layout instead of hugging it next to the text.
2. **"New chat" tooltip now renders on top** (`@align='top-right'`, matching upstream's `cds-icon-button align="top-right"`) instead of a hardcoded `bottom` override - this was a real component bug, not a demo issue.
3. **The dropdown-trigger ("Options") tooltip now has its dark background.** Root cause: `docs-app`'s `theme-support.gts` inlines several components' astroturf-generated CSS into each shadow-wrapped demo so their styles apply inside the shadow root - but `?inline` on a `.module.scss` import returns the CSS-Modules classname map, not compiled CSS text, so those interpolations were a silent no-op (confirmed this is a pre-existing gap affecting 5 other components too, filed as a follow-up rather than expanding this PR's scope). Fixed this tooltip's rule specifically by writing it against the real classname property.

Full checklist green: `pnpm exec glint`, addon `build:js`, addon `lint` (0 errors), full `test-app` suite 1032/1032 (only the pre-existing, already-documented `Grid` snapshot env-flake elsewhere, unrelated). One new regression test added for the icon-sibling structure. Details in AGENTS.md's "Porting Carbon AI Chat" section.

patricklx:
Fixed: the history-item overflow menu ("Rename"/"Delete") could open past the right edge of the history panel.

Root cause: `OverflowMenu` is built on `ember-basic-dropdown`, whose `horizontalPosition` default (`'auto'`) picks whichever side fits the *viewport*. Since the trigger sits flush against the right edge of the (usually much narrower) history panel, `'auto'` almost always resolves to `'left'`, growing the menu rightward out of the panel instead of staying within it.

Fix: added a `@horizontalPosition` passthrough arg to `OverflowMenu` and set it to `'right'` on `ChatHistoryPanelItem`'s menu, so it opens toward the panel's interior (left of the trigger) instead.

Added a regression test asserting the dropdown gets the `ember-basic-dropdown-content--right` state class (rather than a raw pixel-position assertion - confirmed empirically that the QUnit rendering-test harness's own container/coordinate setup makes ember-basic-dropdown's absolute-pixel math unreliable to assert against directly, unrelated to this component). Verified the test fails without the fix (`--left`) and passes with it.

Full checklist: `pnpm exec glint`, addon `build:js`, addon `lint` (0 errors) all clean; full `test-app` suite 1030/1033 (only the pre-existing, already-documented `Grid` white/dark-theme snapshot env-flake).

